PR #12, opened 2019-10-22
refactor opcode

quininer:
solve #11


quininer:
I have a better idea and I will reimplement it later.

PR #16, opened 2019-11-08
Format code using 'cargo fmt' & add rustfmt.toml

Atul9:


quininer:
Thanks for your PR, but I don't want rustfmt right now.

PR #30, opened 2020-06-06
Use mmap64 instead of mmap

Reisz:
Small change to allow building the crate on 32-bit targets.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -21,9 +21,9 @@ pub struct Mmap {
 }
 
 impl Mmap {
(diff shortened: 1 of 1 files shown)


quininer:
I think it might be better to use `libc::off_t` instead of `i64`?

Reisz:
That would cause a bitcast from `u32` to `i32` when performing `as _`. But that could be prevented by making the constants `i32` and using `.into()` instead.

Alternatively, the bitcast could also be ignored because none of the constants are big enough to wrap.

Which one do you prefer?

quininer:
I think rustc will do basic constant propagation, simple cast is good.

quininer:
Thank you!

PR #31, opened 2020-07-30
Allow access to the flags value of CQEs

DaGenix:
Access to the result is already available. The flags value isn't, however. I can't see any reason it shouldn't be - and access to the flags value is necessary to use the buffer selection mechanism.


quininer:
In the early days flags were of no use, and I am not sure of best way to expose it. It looks good now, thanks.

PR #32, opened 2020-07-30
Support setting buf_group parameter for Read and Recv opcodes

DaGenix:
I'm not quite sure where it makes the most sense to expose this value - on the Opcode or on the Entry. With this PR, I expose it as an option on the Opcodes.


quininer:
Opcode is good, thanks.

quininer:
I will merge the `linux-5.8` branch and release a new version at the weekend.

PR #34, opened 2020-07-30
sys: Expose the module as public

### LINKED ISSUE #33: Change sys module from private to public
The `sys` module defines several constants (syscalls, operations, etc...), and it would be useful from the caller's perspective, to be able to access them.

sboeuf:
By making the sys module public, we can let the caller rely on some very
useful constants.

Fixes #33 

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,7 @@
 
 #[macro_use]
 mod util;
-mod sys;
+pub mod sys;
 mod register;
 mod submit;
 pub mod squeue;


sboeuf:
@DaGenix @quininer let me know if this is something you might consider for the crate.
Also, the CI is failing, but I don't think this is related to this PR.

quininer:
@sboeuf This is an unrelated error and we can ignore it. :D

PR #41, opened 2020-11-25
derive Debug, Clone, and Copy where appropriate

dekellum:
Initially, for (transparent) types `Fd` and `Fixed`. 

Note: There may be more places this would be useful, and I'll add those here if it remains open, or in new PRs otherwise.  Thanks!


quininer (approved):
LGTM

quininer:
thank you

PR #45, opened 2020-12-23
Update kernel API to 5.10

quininer:


akiradeveloper:
@quininer 

Hello, can I ask you a simple question?

This fix means users have to kernel 5.10 instead of 5.1 to use tokio/io-uring?

quininer:
@akiradeveloper No, it just adds some new API. The kernel API is backward compatible.

akiradeveloper:
@quininer Thank you.

PR #47, opened 2021-01-03
Improve docs

Koxiaet:
It should make the library easier to understand :)

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -177,8 +177,8 @@ impl Drop for AvailableQueue<'_> {
 }
 
 impl Entry {
-    /// If successful, the number of bytes transferred. If an error occurred, this is a negative
-    /// errno code.
+    /// The operation-specific result code. For example, for a [`Read`](crate::opcode::Read)
+    /// operation this is equivalent to the return value of `read(2)` system call.
     pub fn result(&self) -> i32 {
         self.0.res
     }
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -178,7 +178,7 @@ impl Drop for AvailableQueue<'_> {
 
 impl Entry {
     /// The operation-specific result code. For example, for a [`Read`](crate::opcode::Read)
-    /// operation this is equivalent to the return value of `read(2)` system call.
+    /// operation this is equivalent to the return value of the `read(2)` system call.
     pub fn result(&self) -> i32 {
         self.0.res
     }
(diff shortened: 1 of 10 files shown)


quininer:
Thank you so much!

src/opcode.rs
```diff
@@ -48,46 +48,56 @@ mod sealed {
...
     }
 
-    /// Opaque types, you should use `libc::epoll_event` instead.
+    /// Opaque types, you should use [`epoll_event`](libc::epoll_event) instead.
     #[repr(C)]
     pub struct epoll_event {
         _priv: (),
     }
 
+    /// A file descriptor that has not been registered with io_uring. This can be slower than using
```
quininer, line 73: > This can be slower than using [`Fixed`].

I suspect this is not true, maybe we can delete this.

quininer (changes requested):

src/cqueue.rs
```diff
@@ -160,19 +177,23 @@ impl Drop for AvailableQueue<'_> {
 }
 
 impl Entry {
-    /// Result value
+    /// If successful, the number of bytes transferred. If an error occurred, this is a negative
```
quininer, line 180: > If successful, the number of bytes transferred. If an error occurred, this is a negative errno code.

I think the meaning of result depends on opcode.

quininer (changes requested):

src/opcode.rs
Koxiaet: The manpage for `io_uring_register` says that registering file descriptors "greatly reduces per-I/O overhead", so I took that to mean that using registered file descriptors would be faster. Maybe we could make it more specifically about overhead though.

quininer: We can mention in `Fixed` that it will reduce some overhead in some cases.

src/cqueue.rs
dekellum: Yes, so for example I'm using `opcode::Openat2` where the result value, when positive, is the opened file descriptor.  BTW: This raises the question of if Entry should be generic over opcode with some associate types or the like.  It would seem better if for, `opcode::Openat2` result was `Result<opcode::types::Fd, io::Error>` but maybe that necessitates some higher level layer?

quininer: The `io-uring` is low-level abstraction and should not do too much.

quininer (approved):
LGTM

quininer:
Thanks again!

PR #52, opened 2021-01-09
Examples: Fix Token leak in tcp_echo example

StephanDollberg:
We were allocating a new Token for the Recv op but never clean up the
Poll token.

Instead, just reuse the token as done on Recv and Write.


quininer:
Thank you!

PR #55, opened 2021-01-17
Replace mmap_offset! with regular code

Koxiaet:
Small change because I think regular code is slightly easier to understand than macros.


quininer:
Thank you!

Issue #46, opened 2020-12-24
Remove concurrent module ?


quininer:
In practice, concurrent sq is useful, while concurrent cq is almost useless. we can split the two and provide concurrent sq and non-concurrent cq to avoid some atomic overhead.

And, I hope to implement lock-free squeue.

Koxiaet:
I have an idea for how to remove the lock in the concurrent sqe. It still involves waiting, but it should be slightly cheaper. We store a second tail atomic, `reserved_tail`, that stores the ring buffer index up to which there are entries reserved. To push to the sq, threads will:

1. Increment `reserved_tail` atomically.
2. Write their entries to the slots that have been reserved.
3. While attempting to CAS `tail` from the old value of `reserved_tail` to the new value of `reserved_tail` fails, wait on `head`'s futex.
4. Wake `head`'s futex.

Would this work?

quininer:
@Koxiaet  It sounds good, maybe the performance will be better than the lock under light load.

Would you like to try it?

Koxiaet:
@quininer Yes, I'll see if I can implement it.

PR #56, opened 2021-01-17
Add push_multiple

Kestrer:
Currently concurrent io_uring instances can't really be used to send linked SQEs because another thread could call `push` in between one thread calling `push` twice. This PR adds a new function, `push_multiple`, to allow pushing multiple SQEs atomically.

Since it's not easy to pass in an arbitrary number of owned values to a function in Rust yet, I decided that it would be better to make `squeue::Entry` `Copy` so that it wouldn't be inconsistent to take both `Entry` and `&[Entry]`. ~~And since we no longer need to return back ownership of the `Entry` when the queue is full, the push functions return `bool`s instead of `Result<(), Entry>` - although they could also return `Result<(), ()>`.~~ Actually, after further thought I have decided to use `Result<(), ()>` because it has more obvious semantics and it is `#[must_use]`.

### DIFF
--- a/examples/tcp_echo.rs
+++ b/examples/tcp_echo.rs
@@ -40,7 +40,7 @@ impl AcceptCount {
 
     pub fn push_to(&mut self, sq: &mut squeue::AvailableQueue) {
         while self.count > 0 {
-            if !unsafe { sq.push(self.entry) } {
+            if unsafe { sq.push(self.entry) }.is_err() {
                 break;
             }
             self.count -= 1;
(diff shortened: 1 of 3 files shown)


quininer:
I want to put it aside for now, until we decide on [the future of concurrent module](https://github.com/tokio-rs/io-uring/issues/46).

quininer:
Unless the kernel supports it, it is difficult for us to provide meaningful concurrent mod.

[This PR](https://github.com/tokio-rs/io-uring/pull/64) will allow developers to lock sq by themselves. maybe one day kernel will support concurrent queues, and then we can bring back concurrent mod. .

quininer:
I'm considering adding this to normal sq and cq, it will have some performance advantages (batch submit see ), and I find it useful when writing links like write + flush.

PR #57, opened 2021-01-18
Use futex in concurrent squeue

Kestrer:
Based on my ideas in #46 - now two threads can push to the squeue in parallel, although it does require synchronization in the end.

### DIFF
--- a/src/concurrent/squeue.rs
+++ b/src/concurrent/squeue.rs
@@ -1,15 +1,12 @@
 use std::sync::atomic;
 
 use crate::squeue::{self, Entry};
-use crate::util::unsync_load;
-
-use parking_lot::Mutex;
 
 /// An io_uring instance's submission queue. This contains all the I/O operations the application
 /// sends to the kernel.
 pub struct SubmissionQueue<'a> {
     pub(crate) queue: &'a squeue::SubmissionQueue,
-    pub(crate) push_lock: &'a Mutex<()>,
+    pub(crate) reserved_tail: &'a atomic::AtomicU32,
     pub(crate) ring_mask: u32,
     pub(crate) ring_entries: u32,
 }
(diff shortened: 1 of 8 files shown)


src/concurrent/squeue.rs
```diff
@@ -67,18 +64,63 @@ impl SubmissionQueue<'_> {
...
+        while (*self.queue.tail)
+            .compare_exchange(
+                previous_reserved_tail,
+                previous_reserved_tail.wrapping_add(1),
+                atomic::Ordering::Release,
+                atomic::Ordering::Relaxed,
+            )
+            .is_err()
+        {
+            libc::syscall(
```
quininer, line 98: I think we shouldn't wait when tail == previous_reserved_tail?

src/concurrent/squeue.rs
```diff
@@ -67,18 +64,63 @@ impl SubmissionQueue<'_> {
...
+                std::ptr::null::<u32>(),
+                // Ignored by FUTEX_WAIt.
+                0_u32,
+            );
         }
 
-        *self.queue.sqes.add((tail & self.ring_mask) as usize) = entry;
-
-        (*self.queue.tail).store(tail.wrapping_add(1), atomic::Ordering::Release);
+        libc::syscall(
```
quininer, line 112: We should add a function that does not depend on libc. like https://github.com/tokio-rs/io-uring/blob/master/src/sys/mod.rs#L20

quininer:
We also need to add a bench to prove that it has a performance advantage over locks.

src/concurrent/squeue.rs
Koxiaet: Yes, my bad - I've fixed that now.

Koxiaet:
So I wrote some improved benchmarks to try and see how this would change performance - but to my dismay this change actually _decreases_ performance in the heavily contended case. The lock-based approach took about 130ns to push a single item, whereas the futex approach takes 190ns! I committed the benchmarks I wrote - so before we close this we should at least check them to make sure they were good benchmarks.

quininer:
The result is a bit regretful. I guess this is because it almost always never go to `futex_wait` when heavily contended, so that it degenerates into a spinlock.

PR #62, opened 2021-01-22
Allow setting squeue Entry Personality

### LINKED ISSUE #58: Support setting the credential ID of SQEs
This can be used to share a ring between separate users or processes.

mxxo:
Fixes #58 by adding an `Entry::personality` method and `Personality` wrapper type. Please let me know if you'd like me to make any changes.

### DIFF
--- a/src/register.rs
+++ b/src/register.rs
@@ -85,6 +85,14 @@ impl Drop for Probe {
     }
 }
 
+/// An id associated with the credentials of the current process. After [registering](crate::Submitter::register_personality)
+/// the id, it can be [passed](crate::squeue::Entry::personality) into submission queue entries
+/// to issue the request with the process' credentials.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Personality {
+    pub(crate) id: u16,
+}
+
 /// An allowed feature of io_uring. You can set the allowed features with
 /// [`register_restrictions`](crate::Submitter::register_restrictions).
 ///
--- a/src/register.rs
+++ b/src/register.rs
@@ -85,14 +85,6 @@ impl Drop for Probe {
     }
 }
 
-/// An id associated with the credentials of the current process. After [registering](crate::Submitter::register_personality)
-/// the id, it can be [passed](crate::squeue::Entry::personality) into submission queue entries
-/// to issue the request with the process' credentials.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub struct Personality {
-    pub(crate) id: u16,
-}
-
 /// An allowed feature of io_uring. You can set the allowed features with
 /// [`register_restrictions`](crate::Submitter::register_restrictions).
 ///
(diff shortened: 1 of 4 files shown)


src/register.rs
```diff
@@ -85,6 +85,14 @@ impl Drop for Probe {
     }
 }
 
+/// An id associated with the credentials of the current process. After [registering](crate::Submitter::register_personality)
+/// the id, it can be [passed](crate::squeue::Entry::personality) into submission queue entries
+/// to issue the request with the process' credentials.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Personality {
```
quininer, line 92: I think it's good to use u16 directly, which seems to be a unnecessary newtype.

src/squeue.rs
```diff
@@ -266,4 +267,10 @@ impl Entry {
         self.0.user_data = user_data;
         self
     }
+
+    /// Set the [`Personality`] of this event.
+    pub fn personality(mut self, personality: Personality) -> Entry {
```
quininer, line 272: for function that have not been tested, I think it is placed behind unstable feature gate.

quininer (changes requested):

src/register.rs
mxxo: You got it, done in 940e80a

src/squeue.rs
mxxo: Thank you, that makes sense (226bef3)

src/submit.rs
```diff
@@ -275,16 +275,14 @@ impl<'a> Submitter<'a> {
...
         let id = execute(
             self.fd.as_raw_fd(),
             sys::IORING_REGISTER_PERSONALITY,
             ptr::null(),
             0,
         )?;
-        Ok(Personality {
-            id: id.try_into().unwrap(),
-        })
+        Ok(id.try_into().unwrap())
```
quininer, line 285: I think we can avoid `unwrap`, because `execute` has already determined that` ret >= 0`.

maybe we should let `execute` return u16.

mxxo: Would you prefer I just use `id as u16` here?

quininer: I believe it makes more sense to make `execute` return u16.

mxxo: Is there the possibility that `execute` returns an i32 that doesn't fit in a u16?

quininer: Currently `execute` never return negative number. If there is a new method that needs to return i32 in future, we can also simply change it without breaking API.

mxxo: Ok, but since `io_uring_register` returns an i32, isn't it possible it could be a positive number larger than `u16::MAX` (65535)? (Sorry if I'm not understanding something here)

quininer: You're right... It's four o'clock in the morning, I might need sleep.

I think `id as u16` is good.

mxxo: Ok, no problem at all! :)

quininer:
Thank you!

PR #69, opened 2021-02-16
Take an &Entry in squeue push

Kestrer:
This provides a more consistent interface between `push` and `push_multiple`. It also removes the need for `.ok()` before `.expect` in many of the tests, as the error type is a printable `Result` already.


quininer:
This looks great, thanks!

PR #70, opened 2021-02-16
Rename Insufficient to PushError

Kestrer:
`PushError` is more inline with Rust error naming conventions. As it's now an error I implemented `Display` and `Error` as well as a couple other basic traits.


quininer:
Thank you!

Issue #61, opened 2021-01-20
Is AvailableQueue necessary?


Kestrer:
Currently we have `squeue::AvailableQueue` and `cqueue::AvailableQueue` that represent a snapshot of the submission queue and completion queue respectively. But I don't think either are necessary: performance-wise, atomic loads and stores are incredibly cheap, and utility-wise there isn't much benefit in freezing the queue at a specific point in time, and if users wish to do that they can just cap it off at the length of the queue manually.

I think it would be a lot simpler if we got rid of them entirely, and made `CompletionQueue` implement `Iterator` and have the `push` method on `SubmissionQueue`.

quininer:
I agree that `AvailableQueue` may not much performance advantage, but I don't think we need to get rid of it.

We can simply implement `push` for `SubmissionQueue` and implement `Iterator` for `CompletionQueue`, which does not conflict with `AvailableQueue`. but so far I haven't seen too many benefits to implement these. because we can always simply call `sq.available().push()` and `cq.available().next()`.

Koxiaet:
I would like to get rid of it because it simplifies the API, and reduces the amount of work users have to do to learn how to use this library. It also reduces the maintainance cost - not having to maintain two sets of `capacity`/`len`/`is_empty`/`is_full` methods and keep their docs and tests in sync etc is a large benefit in my opinion. Also I find it's somewhat of a misnomer, "available queue" doesn't really accurately describe what it does.

quininer:
These methods only have dozens of lines of code in total, and I don't think this is a heavy burden. If it is a burden, I will consider remove the `capacity`/`len`/`is_empty`/`is_full` methods of `SubmissionQueue`/`CompletionQueue`.

I didn't understand what problems the `AvailableQueue` caused.

Koxiaet:
I suppose there aren't any problems, I just would like to have the simplest possible API :smiley:. If we're going to keep the type there's no point in removing those methods, they're useful.

PR #72, opened 2021-02-16
Resolve &mut unsoundness

### LINKED ISSUE #71: Giving out &mut SubmissionQueue and &mut CompletionQueue is unsound
The current API gives out mutable references to the inner `SubmissionQueue` and `CompletionQueue` of the io_uring instance. However this gives the ability for users to replace the queues entirely with a different one, leading to code like this causing unsoundness:

```rust
let mut ring_a = IoUring::new(16)?;
let mut ring_b = IoUring::new(16)?;
mem::swap(ring_a.completion(), ring_b.completion());
drop(ring_a); // ring_a's cq is freed...
let freed_queue = ring_b.completion(); // ...but we can still access it here.
```

I think this is best solved by instead handing out a `CompletionQueueMut`/`cqueue::Mut` type or something like that. I'm working on a fix for this now.

Kestrer:
Fixes #71 and #61.

- The name `Mut` is free to be changed, I chose it to be used with code like `cqueue::Mut`, but I'll admit it doesn't give very good docs and having two types of the same name can be confusing.
- I renamed `submission`/`completion` to `submission_mut`/`completion_mut` as they return `Mut`s, and also it would be nice to have methods that return `&SubmissionQueue`s and `&CompletionQueue`s in the future.
- `Mut::reborrow` is free to be changed, as an alternative it could be called `as_mut`.
- I also removed `AvailableQueue` in favour of using the atomic ops directly, since they're cheap.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,8 +30,8 @@ use util::{Fd, Mmap};
 /// IoUring instance
 pub struct IoUring {
     inner: Inner,
-    sq: SubmissionQueue,
-    cq: CompletionQueue,
+    sq: squeue::Inner,
+    cq: cqueue::Inner,
 }
 
 struct Inner {
@@ -84,7 +84,7 @@ impl IoUring {
         unsafe fn setup_queue(
             fd: &Fd,
             p: &sys::io_uring_params,
-        ) -> io::Result<(MemoryMap, SubmissionQueue, CompletionQueue)> {
+        ) -> io::Result<(MemoryMap, squeue::Inner, cqueue::Inner)> {
             let sq_len = p.sq_off.array as usize + p.sq_entries as usize * mem::size_of::<u32>();
             let cq_len = p.cq_off.cqes as usize
                 + p.cq_entries as usize * mem::size_of::<sys::io_uring_cqe>();
@@ -95,8 +95,8 @@ impl IoUring {
                 let scq_mmap =
                     Mmap::new(fd, sys::IORING_OFF_SQ_RING as _, cmp::max(sq_len, cq_len))?;
 
-                let sq = SubmissionQueue::new(&scq_mmap, &sqe_mmap, p);
-                let cq = CompletionQueue::new(&scq_mmap, p);
+                let sq = squeue::Inner::new(&scq_mmap, &sqe_mmap, p);
+                let cq = cqueue::Inner::new(&scq_mmap, p);
                 let mm = MemoryMap {
                     sq_mmap: scq_mmap,
                     cq_mmap: None,
@@ -108,8 +108,8 @@ impl IoUring {
                 let sq_mmap = Mmap::new(fd, sys::IORING_OFF_SQ_RING as _, sq_len)?;
                 let cq_mmap = Mmap::new(fd, sys::IORING_OFF_CQ_RING as _, cq_len)?;
 
-                let sq = SubmissionQueue::new(&sq_mmap, &sqe_mmap, p);
-                let cq = CompletionQueue::new(&cq_mmap, p);
+                let sq = squeue::Inner::new(&sq_mmap, &sqe_mmap, p);
+                let cq = cqueue::Inner::new(&cq_mmap, p);
                 let mm = MemoryMap {
                     cq_mmap: Some(cq_mmap),
                     sq_mmap,
@@ -173,7 +173,7 @@ impl IoUring {
 
     /// Get the submitter, submission queue and completion queue of the io_uring instance. This can
     /// be used to operate on the different parts of the io_uring instance independently.
-    pub fn split(&mut self) -> (Submitter<'_>, squeue::Mut<'_>, cqueue::Mut<'_>) {
+    pub fn split(&mut self) -> (Submitter<'_>, SubmissionQueue<'_>, CompletionQueue<'_>) {
         let submit = Submitter::new(
             &self.inner.fd,
             &self.inner.params,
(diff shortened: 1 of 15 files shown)


quininer:
iai bench seems to have a performance regression.

before:
```
bench_io_uring
  Instructions:                3588 (No change)
  L1 Accesses:                 5661 (No change)
  L2 Accesses:                   18 (No change)
  RAM Accesses:                  80 (No change)
  Estimated Cycles:            8551 (No change)

bench_io_uring_batch
  Instructions:                2968 (No change)
  L1 Accesses:                 4546 (No change)
  L2 Accesses:                   22 (No change)
  RAM Accesses:                  91 (No change)
  Estimated Cycles:            7841 (No change)
```

after:
```
bench_io_uring
  Instructions:                5900 (+64.43701%)
  L1 Accesses:                10404 (+83.78378%)
  L2 Accesses:                   15 (-16.66667%)
  RAM Accesses:                  60 (-25.00000%)
  Estimated Cycles:           12579 (+47.10560%)

bench_io_uring_batch
  Instructions:                4406 (+48.45013%)
  L1 Accesses:                 7320 (+61.02068%)
  L2 Accesses:                   25 (+13.63636%)
  RAM Accesses:                  96 (+5.494505%)
  Estimated Cycles:           10805 (+37.80130%)
```

src/squeue.rs
```diff
@@ -20,14 +20,11 @@ pub struct SubmissionQueue {
...
-    tail: u32,
-    ring_mask: u32,
-    ring_entries: u32,
-    queue: &'a mut SubmissionQueue,
-}
+/// A type that grants mutable access to an io_uring's [submission queue](SubmissionQueue).
+///
+/// This is necessary to prevent users swapping out the submission queue, which can cause
+/// unsoundness.
+pub struct Mut<'a>(pub(crate) &'a mut SubmissionQueue);
```
quininer, line 27: I think we can keep the name of `SubmissionQueue` and make the original `SubmissionQueue` a private structure. to avoid excessive API changes.

src/lib.rs
```diff
@@ -173,26 +173,26 @@ impl IoUring {
...
-        (submit, &mut self.sq, &mut self.cq)
+        (submit, squeue::Mut(&mut self.sq), cqueue::Mut(&mut self.cq))
     }
 
-    /// Get the submission queue of the io_uring instace. This is used to send I/O requests to the
+    /// Get the submission queue of the io_uring instance. This is used to send I/O requests to the
     /// kernel.
-    pub fn submission(&mut self) -> &mut SubmissionQueue {
-        &mut self.sq
+    pub fn submission_mut(&mut self) -> squeue::Mut<'_> {
```
quininer, line 189: I think it can keep its original name.

src/squeue.rs
```diff
@@ -178,55 +175,15 @@ impl SubmissionQueue {
...
-    /// can be added before the kernel consumes some.
-    #[inline]
-    pub fn is_full(&self) -> bool {
-        self.tail.wrapping_sub(self.head) == self.ring_entries
+impl Mut<'_> {
+    /// Reborrow this mutable accessor to a shorter lifetime.
+    ///
+    /// This can be used to avoid consuming the `Mut` when passing it to functions.
+    #[must_use]
+    pub fn reborrow(&mut self) -> Mut<'_> {
```
quininer, line 185: :heart:

src/squeue.rs
```diff
@@ -178,55 +175,15 @@ impl SubmissionQueue {
...
-                tail: unsync_load(self.tail),
-                ring_mask: self.ring_mask.read(),
-                ring_entries: self.ring_entries.read(),
-                queue: self,
-            }
-        }
-    }
 }
 
-impl AvailableQueue<'_> {
```
quininer, line 197: I still want to keep `AvailableQueue`, which has better performance in theory and is not complicated.

quininer (changes requested):

Kestrer:
One of the reasons I decided to use `submission_mut` and `Mut` is that I would like to support using the queues from shared references to the io_uring instance. For example currently it's not possible to check whether the kernel thread needs waking up (`SubmissionQueue::need_wakeup`) from an `&IoUring`, when it could be if we provided `fn submission(&self) -> &SubmissionQueue` as well as `fn submission(&mut self) -> &mut SubmissionQueue`. 

Also, if we later exposed an extra-`unsafe` API for pushing to an `&SubmissionQueue`, it would allow `ownedsplit` to be implemented entirely using the public API of the library, which is good for code readability and would give users the opportunity of making their own similar data structures.

quininer:
Returning `&SubmissionQueue` sounds easy to be abused. If we want to know whether we need to wake up, we can expose the `sq_need_wakeup` method of `Submitter`.

I would rather avoid interior mutability.

Kestrer:
> Returning `&SubmissionQueue` sounds easy to be abused.

Is it? What in particular are you thinking of?

> we can expose the sq_need_wakeup method of Submitter.

But it's not just this - it is potentially useful to access _all_ the concurrently accessible properties of the sq and cq from a shared reference, and I wouldn't like to duplicate all the methods.

> I would rather avoid interior mutability.

It wouldn't actually require interior mutability I don't think - we have a `*mut` pointer, so the shared-push methods can just write to that exactly as the existing ones do. The code would be identical, it would just take &self instead of &mut self.

quininer:
> it is potentially useful to access all the concurrently accessible properties of the sq and cq from a shared reference, and I wouldn't like to duplicate all the methods.

I really don't think there will be many repetitive methods. And I don't know which scenes need to access them at the same time.

> The code would be identical, it would just take &self instead of &mut self.

This is the definition of interior mutability.

Kestrer:
Ok, this new design incorporates the performance advantages of an `AvailableQueue`, and avoids the complexity of having both `{Submission, Completion}Queue` and `Mut` - there is only `{Submission, Completion}Queue` in the API. I have not implemented the ability to get sq/cq flags without mutable access to the `IoUring`.

quininer (approved):
This looks great, thank you very much!

PR #73, opened 2021-02-17
Store ring_mask and ring_entries inline

Kestrer:
`u32` takes up less space than a pointer and has one less level of indirection. They're immutable so it should not make a difference to functionality.


quininer:
Thank you!

PR #74, opened 2021-02-17
Return slice from CompletionQueue::fill

Kestrer:
This is slightly more helpful to the user, it reduces the amount of `unsafe` they have to use.

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -139,16 +139,16 @@ impl CompletionQueue<'_> {
     #[cfg(feature = "unstable")]
     #[inline]
     pub fn fill<'a>(&mut self, entries: &'a mut [MaybeUninit<Entry>]) -> &'a mut [Entry] {
-        let len = std::cmp::min(self.len(), entries.len()) as u32;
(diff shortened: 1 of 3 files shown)


src/cqueue.rs
```diff
@@ -138,7 +138,7 @@ impl CompletionQueue<'_> {
 
     #[cfg(feature = "unstable")]
     #[inline]
-    pub fn fill(&mut self, entries: &mut [MaybeUninit<Entry>]) -> usize {
+    pub fn fill<'a>(&mut self, entries: &'a mut [MaybeUninit<Entry>]) -> &'a mut [Entry] {
         let len = std::cmp::min(self.len(), entries.len()) as u32;
```
quininer, line 142: This `as u32` seems redundant.

quininer (changes requested):

src/cqueue.rs
Kestrer: You're right, thanks for noticing! :D

quininer (approved):
LGTM

PR #75, opened 2021-02-17
Store references in squeue and cqueue where possible

Kestrer:
This reduces `unsafe` by storing references to shared atomic values instead of raw pointers. It is sound, as the references are always valid while they exist and C code can't assert uniqueness like Rust's `&mut` can.

It also may be possible to soundly replace `*mut sqe` with `&'static [UnsafeCell<sqe>]` and `*const cqe` with `&'static [UnsafeCell<cqe>]`, but I'm not so sure on that one so I left it out.

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -8,17 +8,15 @@ use crate::sys;
 use crate::util::{unsync_load, Mmap};
 
 pub(crate) struct Inner {
-    head: *const atomic::AtomicU32,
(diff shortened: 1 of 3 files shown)


quininer:
I think the fake `&'static T` is as dangerous as raw pointer, and I would rather use raw pointer to remind myself that I need to pay attention.

Kestrer:
Yeah, that's fair :)

PR #76, opened 2021-02-17
Fix more warnings

Kestrer:
This fixes Clippy warnings, as well as enabling `rust_2018_idioms` and `unused_qualifications` which are warnings I quite like.


quininer (approved):
LGTM

PR #78, opened 2021-02-18
Warn missing debug implementations

Kestrer:
This adds `#![warn(missing_debug_implementations)]` to the crate root, and implements `Debug` for lots of types.

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -1,5 +1,6 @@
 //! Completion Queue
 
+use std::fmt::{self, Debug, Formatter};
 #[cfg(feature = "unstable")]
 use std::mem::MaybeUninit;
 use std::sync::atomic;
@@ -22,6 +23,7 @@ pub(crate) struct Inner {
 }
 
 /// An io_uring instance's completion queue. This stores all the I/O operations that have completed.
+#[derive(Debug)]
 pub struct CompletionQueue<'a> {
     head: u32,
     tail: u32,
@@ -67,6 +69,20 @@ impl Inner {
     #[inline]
     pub(crate) fn borrow(&mut self) -> CompletionQueue<'_> {
         unsafe { self.borrow_shared() }
+    }
+}
+
+impl Debug for Inner {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CompletionQueueInner")
+            .field("head", unsafe { &*self.head })
+            .field("tail", unsafe { &*self.tail })
+            .field("ring_mask", &self.ring_mask)
+            .field("ring_entries", &self.ring_entries)
+            .field("flags", unsafe { &*self.flags })
+            .field("overflow", unsafe { &*self.overflow })
+            .field("cqes", &self.cqes)
+            .finish()
     }
 }
 
@@ -215,6 +231,16 @@ impl Entry {
     #[inline]
     pub fn flags(&self) -> u32 {
         self.0.flags
+    }
+}
+
+impl Debug for Entry {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Entry")
+            .field("result", &self.0.res)
+            .field("user_data", &self.0.user_data)
+            .field("flags", &self.0.flags)
+            .finish()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@
 //!
 //! The crate only provides a summary of the parameters.
 //! For more detailed documentation, see manpage.
-#![warn(rust_2018_idioms, unused_qualifications)]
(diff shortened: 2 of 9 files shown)


quininer:
I don't think we need to implement Debug for all types.

src/cqueue.rs
```diff
@@ -70,6 +72,20 @@ impl Inner {
     }
 }
 
+impl Debug for Inner {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CompletionQueueInner")
+            .field("head", unsafe { &*self.head })
```
quininer, line 78: This is unsound.

src/lib.rs
```diff
@@ -334,6 +336,57 @@ impl Builder {
     }
 }
 
+impl Debug for Builder {
```
quininer, line 339: I suggest that we add Debug implementations as needed. The value of Builder is set by user, and I think it is rarely needed to print it.

src/lib.rs
```diff
@@ -428,6 +481,26 @@ impl Parameters {
     }
 }
 
+impl Debug for Parameters {
```
quininer, line 484: This might make sense.

src/squeue.rs
```diff
@@ -142,6 +143,20 @@ impl Inner {
     }
 }
 
+impl Debug for Inner {
```
quininer, line 146: I think users should access values through methods such as `.len()`, not through Debug.

src/util.rs
```diff
@@ -61,6 +62,7 @@ impl Drop for Mmap {
 }
 
 /// An owned file descriptor.
+#[derive(Debug)]
```
quininer, line 65: The user can print `fd.0`.

quininer (changes requested):

src/util.rs
Kestrer: This impl was here so I could `#[derive(Debug)]` on types that contain `Fd`s, avoiding manual implementations in many cases.

src/squeue.rs
Kestrer: This impl is necessary so that we can `#[derive(Debug)]` on the `IoUring` itself. Are you saying I should instead print out the length of the squeue in the `Debug` impl?

src/lib.rs
Kestrer: OK, I can remove this.

src/util.rs
quininer: I don't recommend users to use `types::Fd` directly, it is basically only used for opcode construction.

src/squeue.rs
quininer: Is there any practical use case need to print IoUring?

Kestrer: Well, debugging. There are all sorts of bugs that can occur with an API like this, and we never know what information users might need to help resolve them. Also, there is a [guideline that everything must implement Debug](https://rust-lang.github.io/api-guidelines/debuggability.html#all-public-types-implement-debug-c-debug).

src/util.rs
Kestrer: This is `util::Fd` not `types::Fd`. `util::Fd` is contained by `IoUring` and allows it to derive `Debug`. `types::Fd` already implemented `Debug` before this change.

src/squeue.rs
quininer: The user can manually access any useful information without going through the Debug method.

In production, I encountered a lot of problems caused by abuse of Debug, so I was cautious about implementing Debug.

src/util.rs
quininer: Sorry, I read it wrong. :'(

src/squeue.rs
Kestrer: That's true, but `dbg!(&uring);` is a lot easier than having to know the fields individually - especially when debugging we want to make it easy. Problems by abusing `Debug` can be avoided with the [`clippy::use_debug`](https://rust-lang.github.io/rust-clippy/master/index.html#use_debug) lint, so while it does require users to manually add that it is something they should be doing in production.

quininer:
My opinion is that if we find that a certain type does need Debug, it will not be too late to add it at that time.

Kestrer:
But uses of the implementations of `Debug` will very rarely actually make it into commits, since it's for debugging.  So either we end up having commits that randomly add implementations of `Debug`, or we have to write out an implementation of `Debug` every time we want to debug it, and then remove it again when it's fixed, which is a lot of unnecessary work.

Also, it can be useful for users to have the `Debug` implementations, and we can't really expect them to contribute just to add it.

quininer:
First of all, I am a user of the library, and I will know which types are worth implementing Debug.

I would accept simple or useful Debug implementations, but the implementation of IoUring looks too complicated and there are no real use cases.

Kestrer:
Ok, in that case I'll close this.

PR #79, opened 2021-02-23
Remove reborrow on sq and cq

Kestrer:
Because this method copies over the head/tail to a new queue, the original queue becomes outdated and useless, and pushing to it might result in overwriting sqes or at worst reading into uninitialized memory.


quininer:
Thank you! I am planning to remove it.

PR #80, opened 2021-02-23
Add IoUring::builder

Kestrer:
This makes the `Builder` more discoverable and follows Rust conventions. See [Matklad's blog post](https://matklad.github.io/2020/08/12/who-builds-the-builder.html) for more rationale. It's a small enough change that I don't think it needs to be `unstable`.


quininer:
Thank you!

PR #81, opened 2021-02-23
Make setup_attach_wq safer by accepting &IoUring

Kestrer:
By accepting an `&IoUring` instead of an untyped file descriptor we reduce the chance that users will make the mistake of passing in a non-`io_uring` file descriptor. The old, "raw" functionality is still available through `setup_attach_wq_raw`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -296,9 +296,22 @@ impl Builder {
         self
     }
 
+    /// Share the asynchronous worker thread backend of this io_uring with the specified exisiting
+    /// [`IoUring`] instance instead of creating a new thread pool.
+    ///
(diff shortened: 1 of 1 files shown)


quininer:
I prefer to add example docs. The type-safe API is good, but I don't think it is very necessary here.

Kestrer:
The other benefit is being able to use the method without importing `std::os::unix::io::AsRawFd` which is quite useful for ergonomics.

An alternative is to accept an `&impl AsRawFd`, how would you feel about that?

quininer:
I think this is a cold method, so ergonomic optimization is not necessary for this.

Kestrer:
Ok, I will close this then. I guess the extra import isn't _too_ bad.

Issue #60, opened 2021-01-20
Move unsafety to entry creation


Koxiaet:
`io_uring` requires at least one unsafe function; currently this is `push` on the submission queue. However moving this unsafety to the creation of the `squeue::Entry`s (in the opcode module) has a few benefits:
- Some operations such as `Nop` can be made fully safe.
- The unsafety is closer to where users will actually have to audit.
- The safety docs will be more helpful. For example, for `Timeout` writing "`timespec` must be a valid pointer for the lifetime of this entry" in `Timeout::new` is clearer than writing "the entry must be valid for the entire operation" in `push`.

quininer:
I don't think this is useful. The only safe entry is `Nop`, `{Poll,Timeout}Remove` and `AsyncCancel`, and `Nop` has almost no meaning in practice.

This crate is a low-level library, and developer must have sufficient knowledge to use it. Just like pointer, it is safe to build it, but it is unsafe to use it.
It is useful to add security docs about `timespec`, and I believe it is worth it.

Move unsafe code is not enough to prevent abuse. for safety abstraction, I think it should be implement in another library. I am planning to pick it up again recently. https://github.com/quininer/ritsu/issues/17

Koxiaet:
> Just like pointer, it is safe to build it, but it is unsafe to use it.

I see, that's a good reason it should stay the way it is. I suppose the entries in this API are more like pointers than they are like references.

PR #98, opened 2021-08-05
Change raw pointers to references

LinkTed:
I am changing the Opcode structs to be more rust like by changing the raw pointers to references.

This is a step to make it more secure by using the borrow checker of rust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ fn main() -> io::Result<()> {
     let mut ring = IoUring::new(8)?;
 
     let fd = fs::File::open("README.md")?;
-    let mut buf = vec![0; 1024];
+    let mut buf = vec![0; 16];
 
     let read_e = opcode::Read::new(types::Fd(fd.as_raw_fd()), buf.as_mut_ptr(), buf.len() as _)
         .build()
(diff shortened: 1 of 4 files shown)


quininer:
Please note that Entry is deliberately designed as a raw pointer interface. io-uring is a low-level library, so it provides as primitive abstraction as possible. see https://github.com/tokio-rs/io-uring/issues/60

If you want a safe high-level abstraction, you can try [tokio-uring](https://github.com/tokio-rs/tokio-uring).

LinkTed:
Okay, we can use lifetimes for the Entry struct too. Is there any reason, why not using the `slice` or `IoSlice(Mut)` for the opcode structs? Even, if we currently cannot vanish any unsafe code/API.

quininer:
> This is a step to make it more secure by using the borrow checker of rust.

This is not helpful for io-uring, because borrowing cannot guarantee the lifetime of buffer passed to kernel.

In this scenario, using references does not bring any security advantages over raw pointers.

LinkTed:
Yes, in my opinion, it guarantees that the buffer is not dropped before it is submitted to the kernel.

quininer:
In use, Opcode is usually quickly converted to Entry and pushed into the queue. This guarantee doesn't seem to make much sense.

Opcode is just a builder to ensure that users will not set the wrong flag.

Kestrer:
Using a lifetime might trick the user into thinking that this is an at all safe API, or that the lifetime represents the actual lifetime of the data (and not the lifetime between creating the data and passing a reference to it to the kernel). With raw pointers, it is made much more explicit that this is an `unsafe` API which can easily cause use-after-frees, so users would be more careful.

PR #103, opened 2021-08-14
Stable submit_with_args

### LINKED ISSUE #90: Wait for completion
liburing has a way to block until completion events come in. I'd be nice to have that functionality in this library. Some like https://unixism.net/loti/ref-liburing/completion.html#c.io_uring_wait_cqe

quininer:
close #90


quininer:
well, I forgot that it does not exist in 5.10 kernel. We must first update kernel to 5.11 or above to pass test.

PR #112, opened 2021-09-16
Correct error in documentation 

### LINKED ISSUE #111: Clearify documentation
The documentation for the Read opcode says it's equivalent to `read(2)` but I believe that's incorrect as you have to specify `offset = -1` to get the behavior for `read(2)`. The docs should reflect that.

zethra:
Fixes #111

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -184,6 +184,13 @@ impl ExactSizeIterator for CompletionQueue<'_> {
 impl Entry {
     /// The operation-specific result code. For example, for a [`Read`](crate::opcode::Read)
     /// operation this is equivalent to the return value of the `read(2)` system call.
+    ///
+    /// Many system calls that have io-uring equivalents use `errno` to return error values.
+    /// `errno` is not used with io-uring as it's an async interface. Instead the result is set to
+    /// `-errno` in these cases. This can be converted to an [`std::io::Error`] by
+    /// ```rust
+    /// std::io::Error::from_raw_os_error(-entry.result())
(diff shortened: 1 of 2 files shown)


zethra:
I've added a second commit clarifying how error codes are handled.

src/opcode.rs
```diff
@@ -745,7 +745,9 @@ opcode!(
 );
 
 opcode!(
-    /// Read from a file descriptor, equivalent to `read(2)`.
+    /// Read from a file descriptor, mostly equivalent to `pread(2)`.
```
quininer, line 748: Maybe we can follow the latest manpage of liburing.

src/cqueue.rs
```diff
@@ -184,6 +184,9 @@ impl ExactSizeIterator for CompletionQueue<'_> {
 impl Entry {
     /// The operation-specific result code. For example, for a [`Read`](crate::opcode::Read)
     /// operation this is equivalent to the return value of the `read(2)` system call.
+    /// Except, in an error case many system calls would return `-1` and set errno.
```
quininer, line 187: I believe that the meaning of `result` is not universal, and the meaning of `result` of each opcode may be different. We should mention this here and describe the meaning of `result` in each opcode document.

quininer (changes requested):

zethra:
I submitted an issue to liburing as well and they updated their docs so I'll update this PR what the language they used.

zethra:
I updated the comments based on the revised liburing docs.

Issue #116, opened 2021-11-30
Provide interfaces to prepare SQE in place


jiangliu:
Currently the way to submit an SQE is:
1) prepare an opcode object `op`
2) convert the opcode into SQE by op.build()
3) submit the SQE to SQ by sq.push()/push_multiple()

It would be great to provide interfaces to prepare SQE in place to reduce one memory copy by:
1) prepare an opcode object `op`
2) check space left on the SQ and get available SQE by `let sqe: &mut Entry = sq.prepare()`
3) Prepare the SQE inplace by `op.prepare(sqe)`
4) commit the sqe by sq.commit()

quininer:
I think memory copy is not a big overhead, because sqe only has 64bytes. see nop bench, the io_uring crate is not at a performance disadvantage because of this.

but I don't reject this idea, it requires some refactor for opcode mod.

jiangliu:
> I think memory copy is not a big overhead, because sqe only has 64bytes. see nop bench, the io_uring crate is not at a performance disadvantage because of this.
> 
> but I don't reject this idea, it requires some refactor for opcode mod.

It would be great if we could also introduce `pub trait Opcode` too, then we could pass opcode as generic types such as "pub fn prepare<T: Opcode>(...)".

And I will give a try to prepare some patches:)

quininer:
This is not enough, I hope to be able to specify some initialization parameters at same time.

I would consider adding a `PrepareEntry` type and each opcode can initialize it.

```rust
impl SubmissionQueue<'_> {
    fn prepare(&mut self) -> PrepareEntry<'_>;
}

struct PrepareEntry<'a> {
    ptr: *mut io_uring_sqe,
    tail: &'a mut u32
}

impl PrepareEntry<'_> {
    fn commit(self);
}

impl Timeout {
    fn init_to(entry: &mut PrepareEntry, timespec: *const types::Timespec) -> &mut Timeout;
}
```

The current opcode always constructs a separate struct and then constructs sqe type, we need to make it directly modify sqe type.

quininer:
I started to make opcode support inplace init in [the branch](https://github.com/quininer/io-uring/tree/inplace-opcode). This seems to have some performance degradation for ZST like `Nop`, but it should have little effect in production.

quininer:
[...]

Issue #115, opened 2021-10-20
"cargo run --package io-uring-test" compiling failed


ShuochengWang:
env: ubuntu 18.04 docker image with kernel 5.11
version: latest master
report error: 
In function io_uring_test::tests::fs::test_statx': /root/docker/ngo/deps/io-uring/io-uring-test/src/tests/fs.rs:316: undefined reference to statx'

cargo run --package io-uring-test --verbose
       Fresh cfg-if v1.0.0
       Fresh ppv-lite86 v0.2.14
       Fresh sc v0.2.5
       Fresh remove_dir_all v0.5.3
       Fresh bitflags v1.3.2
       Fresh once_cell v1.8.0
       Fresh libc v0.2.104
       Fresh anyhow v1.0.44
       Fresh getrandom v0.2.3
       Fresh socket2 v0.4.2
       Fresh io-uring v0.5.2 (/root/docker/ngo/deps/io-uring)
       Fresh rand_core v0.6.3
       Fresh rand_chacha v0.3.1
       Fresh rand v0.8.4
       Fresh tempfile v3.2.0
   Compiling io-uring-test v0.1.0 (/root/docker/ngo/deps/io-uring/io-uring-test)
[...]

PR #119, opened 2021-12-01
enhance SubmissionQueue to support generate SQE in place

jiangliu:
Enhance io_uring with new interfaces to support prepare SQE in place. These new interfaces pass argument by reference instead of by value, which then could be used to support [ringbahn](https://github.com/ringbahn/ringbahn).
1) introduce trait PrepareSQE for opcode
2) change opcode to implement the PrepareSQE trait 
3) add SubmissionQueue::push_command()/push_commands() to generate SQE in place.

I have add some unit test cases/one benchmark case. According to my test result, there's no regression for existing SubmissionQueue::push(). The new SubmissionQueue::push_command() is slightly slower than push() by acceptable.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -15,14 +15,103 @@ pub trait PrepareSQE {
     fn prepare(&self, sqe: &mut sys::io_uring_sqe);
 }
 
-macro_rules! assign_fd {
-    ( $sqe:ident . fd = $opfd:expr ) => {
-        match $opfd {
-            sealed::Target::Fd(fd) => $sqe.fd = fd,
-            sealed::Target::Fixed(i) => {
-                $sqe.fd = i as _;
-                $sqe.flags |= crate::squeue::Flags::FIXED_FILE.bits();
-            }
+trait AsSqeValue {
+    type T;
+    fn as_sqe_value(&self) -> Self::T;
+}
+
+impl AsSqeValue for sealed::Target {
+    type T = i32;
+
+    fn as_sqe_value(&self) -> Self::T {
+        match self {
+            sealed::Target::Fd(fd) => *fd as _,
+            sealed::Target::Fixed(i) => *i as _,
+        }
+    }
+}
+
+impl AsSqeValue for types::FsyncFlags {
+    type T = u32;
+
+    fn as_sqe_value(&self) -> Self::T {
+        self.bits()
+    }
+}
+
+impl AsSqeValue for types::TimeoutFlags {
+    type T = u32;
+
+    fn as_sqe_value(&self) -> Self::T {
+        self.bits()
+    }
+}
+
+impl AsSqeValue for u64 {
+    type T = u64;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self as Self::T
+    }
+}
+
+impl AsSqeValue for i64 {
+    type T = u64;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self as Self::T
+    }
+}
+
+impl AsSqeValue for u32 {
+    type T = u32;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self
+    }
+}
+
+impl AsSqeValue for i32 {
+    type T = i32;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self
+    }
+}
+
+impl AsSqeValue for u16 {
+    type T = u16;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self
+    }
+}
+
+impl<U> AsSqeValue for *const U {
+    type T = u64;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self as _
+    }
+}
+
+impl<U> AsSqeValue for *mut U {
+    type T = u64;
+
+    fn as_sqe_value(&self) -> Self::T {
+        *self as _
+    }
+}
+
+macro_rules! set_fd_flags {
+    ( $sqe:ident, $opfd:expr ) => {
+        if matches!($opfd, sealed::Target::Fixed(_)) {
+            $sqe.flags |= crate::squeue::Flags::FIXED_FILE.bits();
+        }
+    };
+    ( $self:ident . $sqe:ident, $opfd:expr ) => {
+        if matches!($opfd, sealed::Target::Fixed(_)) {
+            $self.$sqe.flags |= crate::squeue::Flags::FIXED_FILE.bits();
         }
     };
 }
@@ -42,7 +131,7 @@ macro_rules! opcode {
         pub struct $name:ident {
             $( #[$new_meta:meta] )*
 
-            $( $field:ident : { $( $tnt:tt )+ } ),*
+            $( $field:ident : { $( $tnt:tt )+ } => sqe $(. $sqe_field:ident)+ ),*
 
             $(,)?
 
@@ -50,15 +139,21 @@ macro_rules! opcode {
 
             $(
                 $( #[$opt_meta:meta] )*
-                $opt_field:ident : $opt_tname:ty = $default:expr
+                $opt_field:ident : $opt_tname:ty = $default:expr => sqe $(. $opt_sqe_field:ident)+
             ),*
 
             $(,)?
         }
 
+        $( #[$outer_sqe:meta] )*
+        pub struct $name_sqe:ident { }
+
         pub const CODE = $opcode:expr;
 
-        fn set_used_fields(&$self:ident, $sqe1:ident : &mut sys::io_uring_sqe) $set_used_fields:block
+        fn set_special_fields(&mut $self:ident, $sqe1:ident : &mut sys::io_uring_sqe) {
+            $( op: $set_op_special_fields:block )?
+            $( op_sqe: $set_op_sqe_special_fields:block )?
+        }
     ) => {
         $( #[$outer] )*
         pub struct $name {
@@ -67,6 +162,11 @@ macro_rules! opcode {
         }
 
         impl $name {
+            /// The opcode of the operation. This can be passed to
+            /// [`Probe::is_supported`](crate::Probe::is_supported) to check if this operation is
+            /// supported with the current kernel.
+            pub const CODE: u8 = $opcode as _;
+
             $( #[$new_meta] )*
             #[inline]
             pub fn new($( $field : $( $tnt )* ),*) -> Self {
@@ -75,11 +175,6 @@ macro_rules! opcode {
                     $( $opt_field: $default, )*
                 }
             }
-
-            /// The opcode of the operation. This can be passed to
-            /// [`Probe::is_supported`](crate::Probe::is_supported) to check if this operation is
-            /// supported with the current kernel.
-            pub const CODE: u8 = $opcode as _;
 
             $(
                 $( #[$opt_meta] )*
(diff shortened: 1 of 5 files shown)


quininer:
Can you describe what problem it solves?

jiangliu:
> Can you describe what problem it solves?

Sorry, I has misused an old version as work base, and found a bug in the old version and it's fixed in the latest code base.
BTW, I have tried to follow the way as you described, but feels it's a little complex to support `push_multiple_opcode()`. So I have slightly changed the design and post it here as RFC.
There are still more changes needed for opcode.rs, I will handle it once we have agreed on the design:)

quininer:
The `RefCell<()>` added in this PR has nothing to do with our previous discussion and destroys the meaning of `borrow_shared`. 

The `borrow_shared` is an unsafe api that requires the caller to ensure safety by themselves. It is used to implement cross-threaded apis such as [owned split](https://github.com/tokio-rs/io-uring/blob/1621748caa0f10c90e5c7341c88fe4495b0c038f/src/ownedsplit.rs).

jiangliu:
> The `RefCell<()>` added in this PR has nothing to do with our previous discussion and destroys the meaning of `borrow_shared`.
> 
> The `borrow_shared` is an unsafe api that requires the caller to ensure safety by themselves. It is used to implement cross-threaded apis such as [owned split](https://github.com/tokio-rs/io-uring/blob/1621748caa0f10c90e5c7341c88fe4495b0c038f/src/ownedsplit.rs).

Sure, I started working by reading all the code and then found something interesting:)
I will split that patch into dedicated thread.
BTW, ownedsplit is great future and definitely helps enabling rust async io with io uring.
I'm working on combing ringbahn and io-uring crates togetther:)

jiangliu:
BTW, I failed to run "cargo run --package io-uring-test" with the latest code with an error:
```
  = note: /uring/target/debug/deps/io_uring_test-af1e753fdc660fe4.3x7eimgxzg68dtpw.rcgu.o: In function `io_uring_test::tests::fs::test_statx':
          /uring/io-uring-test/src/tests/fs.rs:316: undefined reference to `statx'
          collect2: error: ld returned 1 exit status
```
And the toolchain used is "nightly-x86_64-unknown-linux-gnu".
Anything special needed to run the test crate?

quininer:
You may be using a distribution that does not support statx, see [here](https://github.com/tokio-rs/io-uring/issues/115#issuecomment-947570460).

jiangliu:
> The `RefCell<()>` added in this PR has nothing to do with our previous discussion and destroys the meaning of `borrow_shared`.
> 
> The `borrow_shared` is an unsafe api that requires the caller to ensure safety by themselves. It is used to implement cross-threaded apis such as [owned split](https://github.com/tokio-rs/io-uring/blob/1621748caa0f10c90e5c7341c88fe4495b0c038f/src/ownedsplit.rs).

Aha, the RefCell actually helps ownedsplit. Say:
```
{
       let uring = IoUring::new();
       let (submit, sq, cq) = uring.split();
       let sq2 = sq.clone();
       let q1 = sq. submission();
       // With the RefCell changes, following statement should panic, which should be the expected behavior:)
       let q2 = sq2.submission();
}
```

jiangliu:
> > The `RefCell<()>` added in this PR has nothing to do with our previous discussion and destroys the meaning of `borrow_shared`.
> > The `borrow_shared` is an unsafe api that requires the caller to ensure safety by themselves. It is used to implement cross-threaded apis such as [owned split](https://github.com/tokio-rs/io-uring/blob/1621748caa0f10c90e5c7341c88fe4495b0c038f/src/ownedsplit.rs).
> 
> Aha, the RefCell actually helps ownedsplit. Say:
> 
> ```
> {
>        let uring = IoUring::new();
>        let (submit, sq, cq) = uring.split();
>        let sq2 = sq.clone();
>        let q1 = sq. submission();
>        // With the RefCell changes, following statement should panic, which should be the expected behavior:)
>        let q2 = sq2.submission();
> }
> ```

My fault, SubmissionUring does not support Clone yet, so it should be safe.

jiangliu:
Hi @quininer , I have updated the MR and feel it's ready for review:)

quininer:
I missed it and I will look at it on the weekend.

src/squeue.rs
```diff
@@ -240,14 +246,77 @@ impl SubmissionQueue<'_> {
...
+            let sqe = self.next_sqe();
+            opcode.prepare(sqe, *options);
+            self.move_forward();
+        }
+
+        Ok(())
+    }
+
+    #[inline]
+    fn next_sqe(&mut self) -> &mut sys::io_uring_sqe {
```
quininer, line 307: This method and the following `move_forward` should be unsafe, because it may return entry being used by kernel and make kernel use the uninitialized entry.

src/squeue.rs
```diff
@@ -285,6 +354,44 @@ impl Entry {
     }
 }
 
+impl OptionValues {
```
quininer, line 357: The name should be more descriptive.

src/squeue.rs
```diff
@@ -240,14 +246,77 @@ impl SubmissionQueue<'_> {
...
+
+    /// Attempts to push an opcode into the submission queue.
+    /// If the queue is full, an error is returned.
+    ///
+    /// # Safety
+    ///
+    /// Developers must ensure that parameters of the opcode (such as buffer) are valid and will
+    /// be valid for the entire duration of the operation, otherwise it may cause memory problems.
+    #[inline]
+    pub unsafe fn push_command<'a, T: PrepareSQE>(
```
quininer, line 264: As a new interface, it should be unstable.

src/squeue.rs
```diff
@@ -240,14 +246,77 @@ impl SubmissionQueue<'_> {
...
+    #[inline]
+    fn next_sqe(&mut self) -> &mut sys::io_uring_sqe {
+        // Safe because we are accessing an entry in the allocated submission queue.
+        unsafe {
+            &mut *self
                 .queue
                 .sqes
-                .add((self.tail & self.queue.ring_mask) as usize) = *entry;
-            self.tail = self.tail.wrapping_add(1);
+                .add((self.tail & self.queue.ring_mask) as usize)
```
quininer, line 313: Return reference seems to be unsound, maybe we should zero out here, or return a raw pointer.

src/opcode.rs
```diff
@@ -5,10 +5,16 @@
...
 
-use crate::squeue::Entry;
+use crate::squeue::{Entry, OptionValues};
 use crate::sys;
 use crate::types::{self, sealed};
 
+/// Trait to prepare an SQE from an opcode object.
+pub trait PrepareSQE {
+    /// Prepare an SQE from an opcode object.
+    fn prepare(&self, sqe: &mut sys::io_uring_sqe, options: Option<&OptionValues>);
```
quininer, line 15: It should not expose internal types, maybe this trait can not be exposed。

quininer (commented):
I am a little bit dissatisfied with the api proposed by this PR. although it can indeed avoid build `io_uring_sqe`, but it cannot avoid need to build Opcode type once.

benchmarks show that it does not have many performance advantages compared to normal push. I am not sure if this change is worthwhile.

```
normal                  time:   [6.0094 us 6.0865 us 6.1785 us]
                        change: [-0.1695% +1.3807% +3.1320%] (p = 0.11 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe

prepare                 time:   [6.3320 us 6.4818 us 6.6643 us]
                        change: [-8.5613% -5.6659% -2.6755%] (p = 0.00 < 0.05)
                        Performance has improved.
```

src/squeue.rs
jiangliu: Fixed

src/opcode.rs
jiangliu: Fixed

src/squeue.rs
jiangliu: How about "SqeCommonOptions"?

src/squeue.rs
jiangliu: Fixed

src/squeue.rs
jiangliu: fixed

jiangliu:
@quininer I have implemented a version to prepare SQE in place. And take `Readv` as an example, the expanded code are as below:
```
    /// Vectored read, equivalent to `preadv2(2)`.
    pub struct Readv {
        fd: sealed::Target,
        iovec: *const libc::iovec,
        len: u32,
        ioprio: u16,
        offset: libc::off_t,
        rw_flags: types::RwFlags,
    }
    impl Readv {
        /// The opcode of the operation. This can be passed to
        /// [`Probe::is_supported`](crate::Probe::is_supported) to check if this operation is
        /// supported with the current kernel.
        pub const CODE: u8 = sys::IORING_OP_READV as _;
        #[inline]
        pub fn new(fd: impl sealed::UseFixed, iovec: *const libc::iovec, len: u32) -> Self {
            Readv {
                fd: fd.into(),
                iovec: iovec.into(),
                len: len.into(),
                ioprio: 0u16,
                offset: 0i64,
                rw_flags: 0i32,
            }
        }
        #[inline]
        pub const fn ioprio(mut self, ioprio: u16) -> Self {
            self.ioprio = ioprio;
            self
        }
        #[inline]
        pub const fn offset(mut self, offset: libc::off_t) -> Self {
            self.offset = offset;
            self
        }
        /// specified for read operations, contains a bitwise OR of per-I/O flags,
        /// as described in the `preadv2(2)` man page.
        #[inline]
        pub const fn rw_flags(mut self, rw_flags: types::RwFlags) -> Self {
            self.rw_flags = rw_flags;
            self
        }
        #[inline]
        pub fn build(self) -> Entry {
            let mut sqe = sqe_zeroed();
            self.prepare(&mut sqe);
            Entry(sqe)
        }
    }
    impl PrepareSQE for Readv {
        #[inline]
        fn prepare(&self, sqe: &mut sys::io_uring_sqe) {
            sqe.opcode = Readv::CODE;
            sqe.fd = self.fd.as_sqe_value() as _;
            sqe.__bindgen_anon_2.addr = self.iovec.as_sqe_value() as _;
            sqe.len = self.len.as_sqe_value() as _;
            sqe.ioprio = self.ioprio.as_sqe_value() as _;
            sqe.__bindgen_anon_1.off = self.offset.as_sqe_value() as _;
            sqe.__bindgen_anon_3.rw_flags = self.rw_flags.as_sqe_value() as _;
            {
                if match self.fd {
                    sealed::Target::Fixed(_) => true,
                    _ => false,
                } {
                    sqe.flags |= crate::squeue::Flags::FIXED_FILE.bits();
                };
            }
        }
    }
    /// Vectored read, equivalent to `preadv2(2)`.
    #[repr(transparent)]
    pub struct ReadvSqe {
        sqe: sys::io_uring_sqe,
    }
    impl ReadvSqe {
        /// The opcode of the operation. This can be passed to
        /// [`Probe::is_supported`](crate::Probe::is_supported) to check if this operation is
        /// supported with the current kernel.
        pub const CODE: u8 = sys::IORING_OP_READV as _;
        #[inline]
        pub fn prepare(&mut self, fd: sealed::Target, iovec: *const libc::iovec, len: u32) {
            self.sqe.opcode = Self::CODE;
            self.sqe.fd = fd.as_sqe_value() as _;
            self.sqe.__bindgen_anon_2.addr = iovec.as_sqe_value() as _;
            self.sqe.len = len.as_sqe_value() as _;
            self.sqe.ioprio = 0u16.as_sqe_value() as _;
            self.sqe.__bindgen_anon_1.off = 0i64.as_sqe_value() as _;
            self.sqe.__bindgen_anon_3.rw_flags = 0i32.as_sqe_value() as _;
            {
                if match fd {
                    sealed::Target::Fixed(_) => true,
                    _ => false,
                } {
                    self.sqe.flags |= crate::squeue::Flags::FIXED_FILE.bits();
                };
            }
        }
        #[inline]
        pub fn get_mut_sqe(&mut self) -> &mut sys::io_uring_sqe {
            &mut self.sqe
        }
        #[inline]
        pub fn ioprio(mut self, ioprio: u16) -> Self {
            self.sqe.ioprio = ioprio.as_sqe_value() as _;
            self
        }
        #[inline]
        pub fn offset(mut self, offset: libc::off_t) -> Self {
            self.sqe.__bindgen_anon_1.off = offset.as_sqe_value() as _;
            self
        }
        /// specified for read operations, contains a bitwise OR of per-I/O flags,
        /// as described in the `preadv2(2)` man page.
        #[inline]
        pub fn rw_flags(mut self, rw_flags: types::RwFlags) -> Self {
            self.sqe.__bindgen_anon_3.rw_flags = rw_flags.as_sqe_value() as _;
            self
        }
    }
    impl<'a> From<&'a mut sys::io_uring_sqe> for &'a mut ReadvSqe {
        #[inline]
        fn from(sqe: &'a mut sys::io_uring_sqe) -> &'a mut ReadvSqe {
            unsafe { mem::transmute(sqe) }
        }
    }
```

jiangliu:
I have also conduct a round of benchmark on one platform. For the Nop tests, the results are as below(in us):
| version | normal | prepare | prepare_sqe|
| ------ | ------ | ---------| -------- |
| original | 13.782| | |
|| 13.834|||
||13.815|||
|first version | 13.786 | 13.776 ||
|| 13.817 | 13.829 ||
|| 13.789 | 14.143 ||
| second version | 13.779 | 13.818 | 13.966 |
|| 13.704 | 13.754 | 13.939 |
|| 13.734 | 13.825 | 13.931 |

For detailed analysis of the performance result, please refer to to https://github.com/tokio-rs/io-uring/issues/116 

There are slight performance degradation for Nop, but I think the performance behave may changes for actual workloads:)

quininer:
This is the same as I expected. opcode is just an entry builder by design. It should be constructed simply and pushed into sq immediately, so it should always be easy to optimize.

If this PR does not provide a visible performance improvement, I tend not to accept it.
Especially I don't like the API it introduces.

jiangliu:
Thanks for your patience, let's close it because it doesn't bring much values:)

quininer:
Anyway, thank you for trying, thank you. :D

PR #120, opened 2021-12-22
Allow debug printing of the submission queue

zethra:
When debugging programs using io_uring it can be useful to check the
state of the submission queue. To make that easier I've added an
implementation of Debug on Entry and SubmissionQueue.

I'm not 100% sure I implemented this correctly so I'd appreciate a code review. I also added a test that just prints the queue full and then empty. I'm not sure if a different or more comprehensive testing method is desired.

### DIFF
--- a/src/squeue.rs
+++ b/src/squeue.rs
@@ -1,7 +1,7 @@
 //! Submission Queue
 
 use std::error::Error;
-use std::fmt::{self, Display, Formatter};
+use std::fmt::{self, Debug, Display, Formatter};
 use std::sync::atomic;
 
 use crate::sys;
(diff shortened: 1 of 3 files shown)


zethra:
It may also be desirable to implement Debug on other types if this is accepted.

src/squeue.rs
```diff
@@ -297,3 +297,27 @@ impl Display for PushError {
...
+    }
+}
+
+impl Debug for SubmissionQueue<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let mut d = f.debug_list();
+        let mut pos = self.head;
+        while pos != self.tail {
+            let entry =
+                Entry(unsafe { *self.queue.sqes.add((pos & self.queue.ring_mask) as usize) });
```
quininer, line 317: we can get `&Entry` directly, thus avoiding a copy?

quininer (changes requested):
I don’t like adding too much `impl Debug`, but if it helps, I’m not against it.

src/squeue.rs
zethra: Yes, done.

Issue #126, opened 2022-04-07
opcode: offset should be u64 instead of libc::off_t


stefano-garzarella:
`libc::off_t` is `i64` while kernel and liburing uses an `u64` for the offset field in most of the opcodes.

There are some exceptions like `IORING_OP_FILES_UPDATE` and `IORING_OP_FALLOCATE`.

To avoid differences and casting issues (e.g. using `try_into()` on u64 to cast to i64 can fail) we should follow what liburing does and expose `offset` as u64.

stefano-garzarella:
And POSIX seems to confirm that `off_t` shall be signed:
`blkcnt_t and off_t shall be signed integer types.`
https://pubs.opengroup.org/onlinepubs/009696899/basedefs/sys/types.h.html

quininer:
Under the impression that the kernel will always convert it to `loff_t`, but I agree that we should follow liburing's interface.

Anyway this is a break change and we need to keep it until the next version.

stefano-garzarella:
> Under the impression that the kernel will always convert it to `loff_t`, but I agree that we should follow liburing's interface.

Right, actually then it always gets converted to `loff_t` which is signed, mah

> Anyway this is a break change and we need to keep it until the next version.

It's fine, please let me know when I should rebase the PR.

PR #127, opened 2022-04-07
opcode: use u64 type for the offset

stefano-garzarella:
`libc::off_t` is `i64` while kernel and liburing uses an `u64`
for the `offset` field in most of the opcodes.

There are some exceptions like `IORING_OP_FILES_UPDATE` and
`IORING_OP_FALLOCATE`.

To avoid differences and casting issues (e.g. using `try_into()`
on `u64` to cast to `i64` can fail) we should follow what liburing
does and expose offset as `u64`.

Fixes: #126
Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -129,7 +129,7 @@ opcode!(
         len: { u32 },
         ;;
         ioprio: u16 = 0,
-        offset64: libc::off64_t = 0,
+        offset: u64 = 0,
         /// specified for read operations, contains a bitwise OR of per-I/O flags,
         /// as described in the `preadv2(2)` man page.
         rw_flags: types::RwFlags = 0,
@@ -141,7 +141,7 @@ opcode!(
     pub fn build(self) -> Entry {
         let Readv {
             fd,
-            iovec, len, offset64,
+            iovec, len, offset,
             ioprio, rw_flags,
             buf_group
         } = self;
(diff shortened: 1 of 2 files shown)


FrankReh:
@quininer What's your thought on this one? Should we ask for a merge or try it ourselves? Had this already been taken care of with another PR?

quininer:
@stefano-garzarella we will be release 0.6 soon, would you like to rebase it so that it can be merged?

stefano-garzarella:
> @stefano-garzarella we will be release 0.6 soon, would you like to rebase it so that it can be merged?

@quininer yep sure, I'll do ASAP!

src/opcode.rs
```diff
@@ -175,7 +175,7 @@ opcode!(
         len: { u32 },
         ;;
         ioprio: u16 = 0,
-        offset64: libc::off64_t = 0,
+        offset64: u64 = 0,
```
quininer, line 178: Can we change it back to `offset`?

quininer (changes requested):

src/opcode.rs
quininer: I should have confirmed that kernel always accepts u64 offset, i don't know why liburing uses `off_t`. maybe we need to check with @axboe .

It is good to keep it as it is for conservative reasons.

axboe: That said, it'd probably be easier for non-c/c++ if we just made them explicitly 64 bits on any build. I'd be fine with doing that, looks like it's just impacting fallocate and madvise/fadvise.

quininer: Thank you for @axboe advice! I think it is always good to use 64bit

stefano-garzarella: @axboe mainly my doubt was about signed/unsigned because `off_t` is signed, so with this PR we remove `off_t` and use `u64` as liburing.

The only one left is `fallocate` where liburing uses `off_t`, but then it casts it to `u64`. Should we use `u64` as well?

axboe: Yes, I'll change fallocate to just use 64-bit offset/length unconditionally.

quininer (approved):
LGTM

quininer:
Thank you!

src/opcode.rs
```diff
@@ -152,20 +152,13 @@ opcode!(
...
     }
 );
 
-impl Readv {
-    #[inline]
-    pub const fn offset(self, offset: libc::off_t) -> Self {
-        self.offset64(offset as libc::off64_t)
-    }
-}
-
```
Congee, line 168: ummm, what's the reason to remove the setter of Readv.offset? How can a user pass an offset to preadv(2) then?

Congee: Ahhh, never mind, I found the builder pattern implemented by the opcode! macro

PR #134, opened 2022-08-17
Implement I/O safety traits

notgull:
This pull request adds a new `io_safety` feature that uses the `AsFd` trait from Rust 1.63. In addition, the `Fd` utility type is implemented using `OwnedFd` for its niching.

See also: sunfishcode/io-lifetimes#38

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -69,7 +69,7 @@ mod fd {
 #[cfg(not(feature = "io_safety"))]
 mod fd {
     use std::mem;
-    use std::os::unix::io::{AsRawFd, RawFd, IntoRawFd, FromRawFd};
+    use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 
     /// API-compatible with the `OwnedFd` type in the Rust stdlib.
     pub struct OwnedFd(RawFd);
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,7 +1,6 @@
-use std::convert::TryFrom;
-use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::os::unix::io::AsRawFd;
 use std::sync::atomic;
-use std::{io, mem, ptr};
+use std::{io, ptr};
 
 /// A region of memory mapped using `mmap(2)`.
 pub struct Mmap {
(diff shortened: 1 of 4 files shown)


src/sys/sys.rs
```diff
@@ -46,7 +46,6 @@ pub const IORING_TIMEOUT_UPDATE: u32 = 2;
 pub const IORING_TIMEOUT_BOOTTIME: u32 = 4;
 pub const IORING_TIMEOUT_REALTIME: u32 = 8;
 pub const IORING_LINK_TIMEOUT_UPDATE: u32 = 16;
-pub const IORING_TIMEOUT_ETIME_SUCCESS: u32 = 32;
```
quininer, line 49: Do not rollback `sys.rs`.

src/util.rs
```diff
@@ -60,49 +59,119 @@ impl Drop for Mmap {
...
+    use std::os::unix::io::{AsRawFd, RawFd, IntoRawFd, FromRawFd, OwnedFd, AsFd, BorrowedFd};
 
-    #[inline]
-    fn try_from(value: RawFd) -> Result<Fd, Self::Error> {
-        if value >= 0 {
-            Ok(Fd(value))
-        } else {
-            Err(())
+    /// An owned file descriptor.
+    pub struct Fd(pub OwnedFd);
```
quininer, line 69: I would make `io_uring` use `OwnedFd` directly.

quininer (changes requested):

src/util.rs
notgull: What is this project's MSRV? I use the `Fd` wrapper to choose between `OwnedFd` and a version compatible with older versions of Rust. Using `OwnedFd` directly would bump the MSRV to 1.63.

quininer: Don't have an explicit MSRV, but I really don't want to chase a compiler version that's too new.

We still have to provide a fallback impl, but the interface should be consistent with `OwnedFd`. After 1.63 is old enough, we can delete this fallback impl directly.

quininer (approved):
LGTM

quininer (changes requested):
Looks like we also need to fix fmt

quininer (approved):

PR #136, opened 2022-08-31
Decrease required Rust version to fix build on stable Debian

albertofaria:
Unfortunately, I didn't notice that my #133 would increase the crate's required Rust version to 1.55, while stable Debian packages are still on 1.48.

Fix build on Debian by avoiding `std::ptr::addr_of` and `std::mem::MaybeUninit::write`.


Noah-Kennedy:
We should not be deleting tests to make this work. If we can't run the tests on older kernels that's fine.

Noah-Kennedy (approved):
Thanks, LGTM!

PR #138, opened 2022-09-08
Use off64_t instead of off_t in opcodes where appropriate

albertofaria:
`off_t` is a synonym for `i32` on 32-bit platforms, even though io_uring always supports 64-bit offsets and fallocate lengths. `off64_t` is always a synonym for `i64`.

- Add `offset64()` opcode methods taking an `off64_t`, but keep the existing `offset()` methods that take `off_t` to maintain backward compatibility.

- Add `Fallocate::new64()` constructor taking an `off64_t` length, but keep the existing `Fallocate::new()` constructor taking an `off_t` length to maintain backward compatibility.

### DIFF
--- a/io-uring-test/src/tests/fs.rs
+++ b/io-uring-test/src/tests/fs.rs
@@ -143,13 +143,23 @@ pub fn test_file_fallocate<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
             .expect("queue is full");
     }
 
-    ring.submit_and_wait(1)?;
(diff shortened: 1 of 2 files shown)


quininer:
This makes `opcode!` even more complicated. Can we add new `Fallocate64` opcode, and deprecate old opcode?

albertofaria:
v2:

- Add a new `Fallocate64` opcode instead of modifying `Fallocate` to use a 64-bit length.
- Deprecate `Fallocate`.

PR #140, opened 2022-09-13
Frankreh/graceful test fallback

### LINKED ISSUE #137: 0.5.5 io-uring-test fails on slightly older kernels
The io-uring-test uses probe to look for features but with the introduction of the second sizes for the submission queue and completion queue, the test will try to setup all four combinations of ring sizes, without probing that the additional ring sizes are supported first.

If probing for this feature isn't supported before trying to use the feature, perhaps rather than a fatal error, a message indicating the test was skipped because a setup parameter was not supported would keep the entire test binary from seeming to have run into a problem?

The test runs fine on a 5.19 kernel but I ran into this on a 5.15 kernel. Certainly not a blocker - just perhaps confusing for someone else trying to use the io-uring-test package for the first time.

FrankReh:
Fixes #137 

I realize there are lots of ways this could have been addressed.

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -16,9 +16,16 @@ fn main() -> anyhow::Result<()> {
 
     #[cfg(all(feature = "unstable", not(feature = "ci")))]
     {
-        test::<squeue::Entry128, cqueue::Entry>(IoUring::generic_new(entries)?)?;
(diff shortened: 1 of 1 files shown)


Noah-Kennedy (approved):

quininer:
Thank you!

PR #145, opened 2022-09-16
add register_files_sparse

FrankReh:
The ability to register a completely sparse file descriptor table is introduced with the 5.19 kernel.

Also, a register test module is added with a single test for the new register_files_sparse submitter method. Besides testing that the method can be run without error, it tests that unregister_files works by checking that the register form cannot be done twice in a row and it checks that the unregister form cannot be done twice in a row.

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -73,6 +73,9 @@ fn test<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     #[cfg(feature = "unstable")]
     tests::queue::test_batch(&mut ring, &test)?;
 
+    // register
+    tests::register::test_register_files_sparse(&mut ring, &test)?;
+
     // fs
     tests::fs::test_file_write_read(&mut ring, &test)?;
     tests::fs::test_file_writev_readv(&mut ring, &test)?;
--- a/io-uring-test/src/tests/register.rs
+++ b/io-uring-test/src/tests/register.rs
@@ -1,0 +1,70 @@
+use crate::Test;
+use io_uring::{cqueue, opcode, squeue, IoUring};
+
+pub fn test_register_files_sparse<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> anyhow::Result<()> {
+    // register_files_sparse was introduced in kernel 5.19, as was the opcode for UringCmd16.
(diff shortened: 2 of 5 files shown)


src/submit.rs
```diff
@@ -179,6 +179,30 @@ impl<'a> Submitter<'a> {
         .map(drop)
     }
 
+    /// Registers an empty file table of nr_files number of file descriptors. The sparse variant is
+    /// available in kernels 5.19 and later.
+    ///
+    /// Registering a file table is a prerequisite for using any request that
+    /// uses direct descriptors.
+    pub fn register_files_sparse(&self, nr: u32) -> io::Result<()> {
```
quininer, line 187: As a new feature that has not been tested on ci, it needs to be marked unstable.

quininer (changes requested):

src/submit.rs
FrankReh: Will do. Should I fix, squash, and submit a new branch?

quininer: You can push directly to current branch

quininer (approved):

io-uring-test/src/main.rs
```diff
@@ -73,6 +73,9 @@ fn test<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     #[cfg(feature = "unstable")]
     tests::queue::test_batch(&mut ring, &test)?;
 
+    // register
+    tests::register::test_register_files_sparse(&mut ring, &test)?;
```
quininer, line 77: A cfg is needed here to ensure that tests can be performed when unstable is not enabled.

io-uring-test/src/tests/register.rs
```diff
@@ -0,0 +1,70 @@
+use crate::Test;
+use io_uring::{cqueue, opcode, squeue, IoUring};
+
+pub fn test_register_files_sparse<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
```
quininer, line 4: same

quininer (changes requested):

io-uring-test/src/tests/register.rs
FrankReh: When I run *cargo test*, it passes. I guess something is adding the *unstable* feature flag when the test is run, or perhaps all feature flags are added? How do I run the test without the *unstable* feature?

quininer (approved):

quininer:
Thank you!

PR #147, opened 2022-10-04
Add IORING_OP_SEND_ZC opcode

ollie-etl:
Adds support for IORING_OP_SEND_ZC opcode introduced in linux 6.0.

Also exposes `IO_URING_ZQE_F_NOTIF` and `IO_URING_ZQE_F_MORE`, to be able to implement completion handling for zerocopy send, which may return multiple CQE's per submission.


ollie-etl:
Yes: It'll take me a day or so to update the PR. I can run it locally, but I don't think the ci image is the correct kernel version. How is that dealt with?

ollie-etl:
Oops, sorry, I assumed this was the tokio-uring repo

ollie-etl:
@quininer : is that sufficient for testing? Its very basic.

quininer:
Thank you!

PR #149, opened 2022-10-16
Split (Submission/Completion)Queue::sync into 2 different functions

boohba:
The reason behind this PR is that you rarely need to do both operations at once.

As an example, consider the following code:
```rust
let mut cq = unsafe { rt.io_uring.completion_shared() };

if cq.is_empty() {
    rt.io_uring.submit_and_wait(1)
        .expect("submit_and_wait");

    cq.sync();
}

while let Some(cqe) = cq.next() {
    todo!();
}
```
In this case `CompletionQueue::sync` is being used to load new CQE's. Therefore `(*self.queue.head).store(self.head, atomic::Ordering::Release);` is unnecessary. It will be executed in the `CompletionQueue` destructor after CQE's are consumed.

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -111,6 +111,20 @@ impl<E: EntryMarker> CompletionQueue<'_, E> {
         }
     }
 
+    /// Flush any entries consumed in this iterator.
+    #[cfg(feature = "unstable")]
+    #[inline]
+    pub fn flush(&self) {
+        unsafe { (*self.queue.head).store(self.head, atomic::Ordering::Release) };
+    }
+
+    /// Load new entries in the queue if the kernel has produced some entries in the meantime.
+    #[cfg(feature = "unstable")]
+    #[inline]
+    pub fn load(&mut self) {
+        unsafe { self.tail = (*self.queue.tail).load(atomic::Ordering::Acquire); }
+    }
+
     /// If queue is full and [`is_feature_nodrop`](crate::Parameters::is_feature_nodrop) is not set,
     /// new events may be dropped. This records the number of dropped events.
     pub fn overflow(&self) -> u32 {
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -115,14 +115,18 @@ impl<E: EntryMarker> CompletionQueue<'_, E> {
     #[cfg(feature = "unstable")]
     #[inline]
     pub fn flush(&self) {
-        unsafe { (*self.queue.head).store(self.head, atomic::Ordering::Release) };
+        unsafe {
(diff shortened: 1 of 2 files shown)


quininer:
I suspect this has a minimal impact on performance, would you be willing to provide a benchmark of what it does?

boohba:
> I suspect this has a minimal impact on performance, would you be willing to provide a benchmark of what it does?

It might not have a critical impact but it still keeps me up at night. It's an extra atomic operation after all.  
Here's a barebones benchmark:
```rust
fn main() {
    let mut io_uring = io_uring::IoUring::new(16).unwrap();
    let mut cq = io_uring.completion();
    let time = std::time::Instant::now();
    for _ in 0..10000 {
        cq.sync();
    }
    println!("sync: {}", time.elapsed().as_nanos());
    let time = std::time::Instant::now();
    for _ in 0..10000 {
        cq.load();
    }
    println!("load: {}", time.elapsed().as_nanos());
}
}
```

```
root@ubuntu-s-1vcpu-1gb-amd-tor1-01:~# ./b
sync: 15560
load: 6820
root@ubuntu-s-1vcpu-1gb-amd-tor1-01:~# ./b
sync: 17730
load: 7780
root@ubuntu-s-1vcpu-1gb-amd-tor1-01:~# ./b
sync: 615071
load: 9030
root@ubuntu-s-1vcpu-1gb-amd-tor1-01:~# ./b
sync: 17530
load: 7770
root@ubuntu-s-1vcpu-1gb-amd-tor1-01:~# ./b
sync: 21330
load: 9930
root@ubuntu-s-1vcpu-1gb-amd-tor1-01:~# ./b
sync: 15400
load: 8450
```

quininer:
Just comparing two atomic operations is not a meaningful benchmark, try adding a nop operation to see if it is faster.
Also you can using criterion to benchmark, like [io-uring-bench](https://github.com/tokio-rs/io-uring/tree/master/io-uring-bench/src).

FrankReh:
@quininer Would you consider merging this, perhaps once the `unstable` feature lines are removed? It doesn't hurt anything and it doesn't appreciably enlarge the API support surface. I don't think it exposes a footgun. The whole API already requires a good understanding for proper use.

quininer:
As I said earlier, I suspect it will have no visible performance impact in production, so I'm inclined not to merge it. I would be happy to change my mind if there is any proof.

boohba:
> As I said earlier, I suspect it will have no visible performance impact in production, so I'm inclined not to merge it. I would be happy to change my mind if there is any proof.

In my humble opinion, even if it doesn't make a significant difference _performance-wise_, it is not an excuse to have unnecessary code in your program. However, the choice is yours.

PR #150, opened 2022-10-22
register_buf_ring and unregister_buf_ring

FrankReh:
Add support for the buffer ring, available since 5.19.


FrankReh:
Oops. Realized I forgot something else.

PR #151, opened 2022-10-22
register_buf_ring and unregister_buf_ring

FrankReh:
Add support for the buffer ring, available since 5.19.

Add functions register_buf_ring and unregister_buf_ring

and make public the sys::io_uring_buf struct as cqueue::io_uring_buf.

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -332,7 +332,3 @@ pub fn buffer_select(flags: u32) -> Option<u16> {
 pub fn more(flags: u32) -> bool {
     flags & sys::IORING_CQE_F_MORE != 0
 }
-
-#[cfg(feature = "unstable")]
-#[allow(non_camel_case_types)]
-pub type io_uring_buf = sys::io_uring_buf;
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -332,3 +332,7 @@ pub fn buffer_select(flags: u32) -> Option<u16> {
 pub fn more(flags: u32) -> bool {
     flags & sys::IORING_CQE_F_MORE != 0
 }
+
+#[cfg(feature = "unstable")]
+#[allow(non_camel_case_types)]
+pub type io_uring_buf = sys::io_uring_buf;
--- a/src/types.rs
+++ b/src/types.rs
@@ -218,3 +218,33 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
         }
     }
 }
+
+#[cfg(feature = "unstable")]
+#[allow(non_camel_case_types)]
+/// io_uring_buf is a stable copy of the sys::io_uring_buf struct.
+/// When using it to write to the ring buffer shared memory,
+/// only the first three fields should be written per entry.
+/// The resv field should not be written, not even as a zero
+/// as the 16 bits on that location are reserved from the first
+/// element to represent the tail index of the ring.
+pub struct io_uring_buf {
+    /// The 64 bit address of the buffer that the uring interface will write to
+    /// and the application can later read from.
+    pub addr: libc::__u64,
+
+    /// The 32 bit length field of the buffer pointed to by addr.
+    pub len: libc::__u32,
+
+    /// The buffer id that will be placed into the CQE when this buffer is taken from
+    /// the ring by the uring interface.
+    pub bid: libc::__u16,
+
+    /// The reserved 16 bits of each buffer ring entry that should not be written to, not even
+    /// zeroed once the buffer ring has been registered. This is the address of the first entry
+    /// where the 16 bit tail index is stored. As entry[0].tail, it is written to by the
+    /// application as the application fills entries in the ring.
+    pub resv: libc::__u16,
+}
+
+/// The offset to the resv field of the io_uring_buf struct.
+pub const RING_BUFFER_TAIL_OFFSET: usize = 14;
--- a/src/types.rs
+++ b/src/types.rs
@@ -220,30 +220,36 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
 }
 
 #[cfg(feature = "unstable")]
-#[allow(non_camel_case_types)]
-/// io_uring_buf is a stable copy of the sys::io_uring_buf struct.
-/// When using it to write to the ring buffer shared memory,
-/// only the first three fields should be written per entry.
-/// The resv field should not be written, not even as a zero
-/// as the 16 bits on that location are reserved from the first
-/// element to represent the tail index of the ring.
-pub struct io_uring_buf {
-    /// The 64 bit address of the buffer that the uring interface will write to
-    /// and the application can later read from.
-    pub addr: libc::__u64,
+pub struct BufRingEntry(sys::io_uring_buf);
 
-    /// The 32 bit length field of the buffer pointed to by addr.
-    pub len: libc::__u32,
-
-    /// The buffer id that will be placed into the CQE when this buffer is taken from
-    /// the ring by the uring interface.
-    pub bid: libc::__u16,
-
-    /// The reserved 16 bits of each buffer ring entry that should not be written to, not even
-    /// zeroed once the buffer ring has been registered. This is the address of the first entry
-    /// where the 16 bit tail index is stored. As entry[0].tail, it is written to by the
-    /// application as the application fills entries in the ring.
-    pub resv: libc::__u16,
+/// An entry in a buf_ring that allows setting the address, length and buffer id.
+#[cfg(feature = "unstable")]
+impl BufRingEntry {
+    /// Sets the entry addr.
+    pub fn set_addr(&mut self, addr: u64) {
+        self.0.addr = addr;
+    }
+    /// Returns the entry addr.
+    pub fn _addr(&mut self) -> u64 {
+        self.0.addr
+    }
+    /// Sets the entry len.
+    pub fn set_len(&mut self, len: u32) {
+        self.0.len = len;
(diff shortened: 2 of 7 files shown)


src/cqueue.rs
```diff
@@ -332,3 +332,7 @@ pub fn buffer_select(flags: u32) -> Option<u16> {
 pub fn more(flags: u32) -> bool {
     flags & sys::IORING_CQE_F_MORE != 0
 }
+
+#[cfg(feature = "unstable")]
+#[allow(non_camel_case_types)]
+pub type io_uring_buf = sys::io_uring_buf;
```
quininer, line 338: does this seem unnecessary?

quininer (changes requested):

quininer:
would you like to add a test?

src/cqueue.rs
FrankReh: I don't understand your question. I'm using that type in tokio-uring to use uring buffer rings.

FrankReh:
> would you like to add a test?

Would a trivial register test like one I added earlier be sufficient? Actually testing buf rings means implementing a buf ring mechanism which comes with all sorts of opinions and I like that io_uring is trying to keep opinions out of its bindings.

quininer:
It is better to have a complete test, but at least a simple test.

src/cqueue.rs
quininer: The layout generated by codegen can change at anytime, we need to make a stable api wrapper around it, just like https://github.com/tokio-rs/io-uring/blob/master/src/types.rs#L96

FrankReh: Thanks for pointing me to an example. That will be easy to do. I'll assume adding to the bottom of the same types.rs will do. Let me know if some other ordering has been established.

FrankReh: > The layout generated by codegen can change at anytime, we need to make a stable api wrapper around it, just like https://github.com/tokio-rs/io-uring/blob/master/src/types.rs#L96

Last night I said okay, I would follow the example you provided. Now I realize I should make sure I understand the reason you are asking because the layout can't be changed, its a struct in the io_uring.h file that defines the memory layout shared between the kernel and the user land. This crate is providing bindings to the syscalls just like the liburing library does and that library also has to expose struct pointers to the userland application that will work with older kernels, the current kernel, and future kernels - so the io_uring.h author can't change the layout on their end either.

I will certainly create this wrapper if it makes the use of the crate easier, but I wanted to first make sure we are both on the same page regarding what is being protected against. The io_uring.h author also knows the kernel interface can't be changed in breaking ways, forwards or backwards.

quininer: My previous word is not correct. io_uring.h guarantees consistent layout, but codegen does not guarantee consistent api for structs.

If the type in io_uring.h changes, for example by adding a new union field, it will break api, so I say make a stable api wrapper.

FrankReh: While the odds that there are a union added are low (my guess), I like your point. I'll create a stable api wrapper.

FrankReh:
I'm trying to build a test that actually uses a buf_ring entry so is taking me a little longer.

FrankReh:
Once the reviews are passed, confirming you will squash the commits into one? Can I offer a patch version bump here too or do you like to keep those separate commits?

FrankReh:
I have a full unit test, but I haven't found a good way to make the file compile when unstable is *not* specified so its not ready for review yet.

FrankReh:
@quininer I have a more full blown unit test for the buf_ring registration and the buf_ring struct that is made public.

Do you know of a way to exclude an entire file from the build when the unstable feature is not given? I may end up putting cfg(not(feature = "unstable")) on many lines of a test file.

quininer:
@FrankReh You can put a `#![cfg(..)]` at the head of file. or `#[cfg(..)] mod your_mod;`

src/cqueue.rs
FrankReh: Please check the new definition.

FrankReh:
Full test has been added too.

src/types.rs
```diff
@@ -218,3 +218,33 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
...
+
+#[cfg(feature = "unstable")]
+#[allow(non_camel_case_types)]
+/// io_uring_buf is a stable copy of the sys::io_uring_buf struct.
+/// When using it to write to the ring buffer shared memory,
+/// only the first three fields should be written per entry.
+/// The resv field should not be written, not even as a zero
+/// as the 16 bits on that location are reserved from the first
+/// element to represent the tail index of the ring.
+pub struct io_uring_buf {
```
quininer, line 230: I hope this to stay in the same style as other types.

like

```rust
pub struct Buf(sys::io_uring_buf);

impl Buf {
    pub fn addr(mut self, addr: u64) -> Buf {
        self.addr = addr;
    }
}
```

quininer (changes requested):

src/types.rs
```diff
@@ -218,3 +218,33 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
...
+
+    /// The buffer id that will be placed into the CQE when this buffer is taken from
+    /// the ring by the uring interface.
+    pub bid: libc::__u16,
+
+    /// The reserved 16 bits of each buffer ring entry that should not be written to, not even
+    /// zeroed once the buffer ring has been registered. This is the address of the first entry
+    /// where the 16 bit tail index is stored. As entry[0].tail, it is written to by the
+    /// application as the application fills entries in the ring.
+    pub resv: libc::__u16,
```
quininer, line 246: This is a reserved byte that may be used in future, it should not be exposed.

src/types.rs
FrankReh: I would have but that scheme does not allow for writing only three out of the four fields. I tried to explain that in the comments. I think I didn't do a good job at that.

src/types.rs
FrankReh: Well, I went with the hardcoded usize of 14. I could use your help in deriving the offset value from the resv field. My comment does say it represents the resv field though.

FrankReh: You were right. The new way I don't need to expose that field after all, while still preserving the length as 16 bytes.

src/types.rs
```diff
@@ -220,30 +220,36 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
...
-    /// The buffer id that will be placed into the CQE when this buffer is taken from
-    /// the ring by the uring interface.
-    pub bid: libc::__u16,
-
-    /// The reserved 16 bits of each buffer ring entry that should not be written to, not even
-    /// zeroed once the buffer ring has been registered. This is the address of the first entry
-    /// where the 16 bit tail index is stored. As entry[0].tail, it is written to by the
-    /// application as the application fills entries in the ring.
-    pub resv: libc::__u16,
+pub struct BufRingEntry(sys::io_uring_buf);
```
quininer, line 223: I think it should be marked `#[repr(transparent)]`, otherwise cast is UB.

FrankReh:
If and when this is accepted, another version bump will be needed.

src/types.rs
```diff
@@ -218,3 +218,40 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
...
+        self.0.bid = bid;
+    }
+    /// Returns the entry bid.
+    pub fn bid(&mut self) -> u16 {
+        self.0.bid
+    }
+}
+
+/// The offset to the resv field of the io_uring_buf struct.
+pub const RING_BUFFER_TAIL_OFFSET: usize = 14;
```
quininer, line 257: Sorry, I thought about it, I'm a little unhappy with this `TAIL_OFFSET`.

I consider an api like this

```rust
impl BufRingEntry {
    pub fn tail(ptr: *const BufRingEntry) -> *const u16 {
        unsafe {
            std::ptr::addr_of!((*ptr).resv)
        }
    }
}

Comment (FrankReh) [reply src/types.rs:257]: So it returns a pointer? Then I use that in the atomic update?

File: src/types.rs [hunk 6]
```diff
@@ -218,3 +218,40 @@ impl<'prev, 'now> SubmitArgs<'prev, 'now> {
...
+
+/// An entry in a buf_ring that allows setting the address, length and buffer id.
+#[cfg(feature = "unstable")]
+impl BufRingEntry {
+    /// Sets the entry addr.
+    pub fn set_addr(&mut self, addr: u64) {
+        self.0.addr = addr;
+    }
+    /// Returns the entry addr.
+    pub fn _addr(&mut self) -> u64 {
```
quininer, line 234: it shouldn't be `&mut` ?

FrankReh: Ops.

src/types.rs
quininer: Yes, you can cast to `AtomicU16`, or just return `*const AtomicU16` here

FrankReh: I get this clippy warning
 ```
error: this public function might dereference a raw pointer but is not marked `unsafe`
   --> src/types.rs:259:34
    |
259 |             std::ptr::addr_of!((*ptr).0.resv)
    |                                  ^^^
    |
    = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
```
Want me to make the whole function unsafe or allow the use to quiet clippy?

FrankReh: I'm going with the former.

quininer: Yes, please mark it as unsafe.

quininer (approved):
LGTM

PR #152, opened 2022-10-28
submit_and_wait: check overflow bit

FrankReh:
The uring interface will set the IORING_SQ_CQ_OVERFLOW bit in sq_flags to indicate some CQE didn't fit in the cqueue and is being held in overflow memory.

The uring interface will only move entries from the overflow memory to the cqueue when it receives the 'enter' syscall and the IORING_ENTER_GETEVENTS bit is set in the flags parameter.

So this condition, of the overflow being set, is a third reason to add this IORING_ENTER_GETEVENTS  bit to the flags passed to 'enter'.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "io-uring"
-version = "0.5.7"
+version = "0.5.8"
 authors = ["quininer <quininer@live.com>"]
 edition = "2018"
 license = "MIT/Apache-2.0"
--- a/src/submit.rs
+++ b/src/submit.rs
@@ -64,6 +64,13 @@ impl<'a> Submitter<'a> {
         }
     }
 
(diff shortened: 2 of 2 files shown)


FrankReh:
This fixes tokio-uring issues [93](https://github.com/tokio-rs/tokio-uring/issues/93) and [147](https://github.com/tokio-rs/tokio-uring/issues/147).

FrankReh:
Should this commit come with a new version number or is that always done as a separate commit?

quininer (approved):
LGTM

quininer:
You can change version number directly in this commit.

FrankReh:
Please squash and merge if you are happy with the 0.5.8 number in Cargo.toml. Thank you.

FrankReh:
Thanks. How long for crates.io index to be updated typically?

PR #155, opened 2022-11-07
Correctly report errors returned from direct-syscall

### LINKED ISSUE #154: direct-syscall feature doesn't report errors correctly
Linux syscalls do not set errno directly - libc is responsible for taking negative return values from the syscall return value and populating errno.

This means that anything depending on errno (like `std::io::Error::last_os_error()`) will not work if you're not using libc.

When using direct-syscall, libc is not used in the `sys` module - so any call to `sys::*` followed by a `last_os_error` will not report errors correctly. One example of this is when calling io_uring_setup https://github.com/tokio-rs/io-uring/blob/6c535d2ce8d05f7a5eb33a82d948e6eb7e19c6f7/src/lib.rs#L175-L182

Cargo.toml
```
[package]
name = "ringer"
version = "0.1.0"
edition = "2021"
[dependencies]
io-uring = { version = "0.5", features = ["direct-syscall"] }
```

src/main.rs
```
fn main() {
    io_uring::IoUring::new(u32::MAX).unwrap();
}
```

Output from `cargo run`:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 0, kind: Uncategorized, message: "Success" }', src/main.rs:2:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If you remove the `direct-syscall` feature you instead get:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', src/main.rs:2:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

You can fix this by handling the conversion yourself, but an alternative is to move from `sc` to https://crates.io/crates/rustix which already does this conversion into its own `io::Result` type - e.g. https://docs.rs/rustix/latest/rustix/io_uring/fn.io_uring_enter.html returns a `rustix::io::Result` (which you can convert this into a `std::io::Result` with `res.map_err(std::io::Error::from)`)

aidanhs:
Fixes #154

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,12 +173,8 @@ impl<S: squeue::EntryMarker, C: cqueue::EntryMarker> IoUring<S, C> {
         }
 
         let fd: OwnedFd = unsafe {
-            let fd = sys::io_uring_setup(entries, &mut p);
-            if fd >= 0 {
-                OwnedFd::from_raw_fd(fd)
-            } else {
-                return Err(io::Error::last_os_error());
-            }
+            sys::io_uring_setup(entries, &mut p)
+                .map(|fd| OwnedFd::from_raw_fd(fd))?
         };
 
         let (mm, sq, cq) = unsafe { setup_queue(&fd, &p)? };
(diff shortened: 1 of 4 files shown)


quininer:
This looks good, would you like to add some regression tests?

PR #157, opened 2022-11-10
update sys.rs to 6.1.0 rc4

quininer:


.gitignore
```diff
@@ -1,3 +1,4 @@
 **/target
 **/*.rs.bk
 Cargo.lock
+include-file.h
```
FrankReh, line 4: why ignore this file? Is this the temporary file you have planned to replace INCLUDE when the env variable is set? I think the temporary file should show up as untracked by git when the person creates it for their temporary setup.

build.rs
```diff
@@ -21,8 +21,18 @@ fn main() {
...
+    if let Some(path) = env::var("BUILD_IO_URING_INCLUDE_FILE")
+        .ok()
+        .filter(|path| !path.is_empty())
+    {
+        builder = builder.header(path);
+    } else {
+        builder = builder.header_contents("include-file.h", INCLUDE);
+    }
+
+    builder
```
FrankReh, line 35: When I have jury-rigged build.rs to build from a local io_uring.h in my home directory, I had to add ".clang_arg("-F.") to this builder chain. With the change to using builder.header(path) above, that's not necessary now?

build.rs
```diff
@@ -21,8 +21,18 @@ fn main() {
...
-        .header_contents("include-file.h", INCLUDE)
+    let mut builder = bindgen::Builder::default();
+
+    if let Some(path) = env::var("BUILD_IO_URING_INCLUDE_FILE")
+        .ok()
+        .filter(|path| !path.is_empty())
+    {
+        builder = builder.header(path);
+    } else {
+        builder = builder.header_contents("include-file.h", INCLUDE);
```
FrankReh, line 32: What is the benefit to replacing all of INCLUDE? Isn't it just the path to io_uring.h that we want to give fine grained control to?

FrankReh (commented):
Just questions.

PR #162, opened 2022-12-13
Drop the "unstable" feature

albertofaria:
This is a proposal to drop the "unstable" feature.

Dependent crates that rely on "unstable" functionality need to restrict the set of supported io-uring versions to a predetermined closed range (e.g. >=0.5.6, <=0.5.8), just in case a future release breaks compatibility. This makes io-uring version incompatibilities in a dependency graph more likely, and can complicate packaging for distributions (see https://bugzilla.redhat.com/show_bug.cgi?id=2124697#c76 for an example).

Let's rely on the fact that we're pre-1.0 to safely try out new APIs instead.

Since we're removing a feature, this will require a bump to 0.6.0. Alternatively, we can keep the feature but make it functionless.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           command: test
           use-cross: true
-          args: --features unstable --target ${{ matrix.target }}
+          args: --target ${{ matrix.target }}
       - uses: actions-rs/cargo@v1
         with:
           command: run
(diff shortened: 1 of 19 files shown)


quininer:
I actually agree with this. I added this feature because io-uring was in a period of rapid development and I wasn't sure if designed interface was suitable. but now development of io-uring crate has slowed down considerably.

I prefer to keep an empty feature.

albertofaria:
v2:

- Keep an unused "unstable" feature on the io-uring crate.

quininer:
Thank you!

PR #163, opened 2022-12-18
Also run CI 'check' job again Rust 1.48

albertofaria:
This should help maintain compatibility with stable Debian, which is still on Rust 1.48.


quininer:
Thank you!

PR #170, opened 2023-01-23
opcode: add support for IORING_OP_MSG_RING

lucab:
This adds a wrapper for the `IORING_OP_MSG_RING` opcode, which was introduced in Linux 5.18.
Notably, this only implements the data message kind (`IORING_MSG_DATA`), without the FD-sending features.

Ref: https://man7.org/linux/man-pages/man2/io_uring_enter2.2.html

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1369,6 +1369,39 @@ opcode!(
     }
 );
 
+// === 5.18 ===
+
+opcode!(
+    /// Send a message (with data) to a target ring.
+    pub struct MsgRingData {
+        ring_fd: { impl sealed::UseFd },
+        result: { i32 },
+        user_data: { u64 },
+        user_flags: { Option<u32> },
+        ;;
+        opcode_flags: u32 = 0
+    }
+
+    pub const CODE = sys::IORING_OP_MSG_RING;
+
+    pub fn build(self) -> Entry {
+        let MsgRingData { ring_fd, result, user_data, user_flags, opcode_flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        sqe.__bindgen_anon_2.addr = sys::IORING_MSG_DATA.into();
+        sqe.fd = ring_fd;
+        sqe.len = result as u32;
+        sqe.__bindgen_anon_1.off = user_data;
+        sqe.__bindgen_anon_3.msg_ring_flags = opcode_flags;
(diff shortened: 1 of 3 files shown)


src/opcode.rs
```diff
@@ -1369,6 +1369,39 @@ opcode!(
...
 
+// === 5.18 ===
+
+opcode!(
+    /// Send a message (with data) to a target ring.
+    pub struct MsgRingData {
+        ring_fd: { impl sealed::UseFd },
+        result: { i32 },
+        user_data: { u64 },
+        user_flags: { Option<u32> },
```
lucab, line 1380: Note for reviewers: the flags-passing logic will only land in Linux 6.3, but I already added it in the signature here in order to skip a semver breaking change later.

FrankReh:
@quininer I think this should be merged. It passes CI, and the tests passed on a 5.15 and a 6.1 kernel for me. Then a tag should be created, based on any other PRs that might go in soon.

FrankReh:
By the way, with the new rustc version there is a new clippy and this repo will need a PR to address them. Can we merge this first and then @quininer or I can submit a PR addressing clippy?

Noah-Kennedy:
@FrankReh I've merged it

quininer:
Sorry, I'm spending my spring festival holiday.

src/opcode.rs
quininer: If it is an optional flag, I think it should go under `;;` ?

like

```rust
    pub struct MsgRingData {
        ring_fd: { impl sealed::UseFd },
        result: { i32 },
        user_data: { u64 },
        ;;
        user_flags: u32 = 0,
        opcode_flags: u32 = 0
    }
```

`Option` seems unnecessary, because if `None` means it is 0.

lucab: It is slightly more complex than that. Looking at the [referenced kernel patch](https://lore.kernel.org/all/20230103160507.617416-1-leitao@debian.org/), there is an explicit new flag `IORING_MSG_RING_FLAGS_PASS` that needs to be set on the SQE.

So effectively `Some(0)` and `None` have different behaviors, because the former also sets a flag (which only newer kernels will understand) while the latter does not (thus it works also on current kernels).

lucab:
@quininer enjoy your time off :tada:
I'm not in a hurry. if @Noah-Kennedy and others have review/merge access, that's enough to sort out details and make progress.

PR #171, opened 2023-01-28
chore: fix clippy 0.1.67 warnings

FrankReh:
These are limited to the io-uring-test source. No clippy warnings from the main library or the benchmark.

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -20,8 +20,7 @@ fn main() -> anyhow::Result<()> {
             Ok(r) => test(r)?,
             Err(e) => {
                 println!(
-                    "IoUring::<squeue::Entry128, cqueue::Entry>::generic_new(entries) failed: {}",
-                    e
+                    "IoUring::<squeue::Entry128, cqueue::Entry>::generic_new(entries) failed: {e}"
                 );
                 println!("Assume kernel doesn't support the new entry sizes so remaining tests being skipped.");
                 return Ok(());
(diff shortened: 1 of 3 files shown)


io-uring-test/src/main.rs
```diff
@@ -20,8 +20,7 @@ fn main() -> anyhow::Result<()> {
             Ok(r) => test(r)?,
             Err(e) => {
                 println!(
-                    "IoUring::<squeue::Entry128, cqueue::Entry>::generic_new(entries) failed: {}",
```
quininer, line 23: Frankly I don't like this new syntax, I'd rather turn it off in clippy.toml.

FrankReh: Isn't the point of `cargo fmt` and `cargo clippy` to remove individual opinions on how the code should be formatted? Having a clean standard to follow, that none of us made, makes it easier for everyone to get along.

quininer: You're right, I usually have no opinion about rustfmt and clippy.

However, clippy is not a reliable standard, and it has a lot of suboptimal suggestions and false positive. we inevitably need to disable some of these lint in production.

Another reason is that this is a new feature introduced in 1.58 and using it will make it can't run tests with old  rustc. and this feature does not bring any real benefit.

FrankReh: Okay, that's a good reason. Thanks for bringing up the 1.58 dependency. I'll close this.

I guess the Debian 1.48 build check isn't made against the io-uring-test binary, since my PR didn't fail CI.

PR #173, opened 2023-01-31
test/register-buf-ring: provide length argument to read op

lucab:
This tweaks the `register_buf_ring` test, passing the length argument for the I/O operation.
The length was previously specified as `0` (which seems to work on kernel 6.0+), but in that way the test failed on some older kernels (failure observed on 5.19):
```
test register_buf_ring
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[]`,
 right: `[84, 104, 101, 32, 113, 117, 105, 99, 107, 32, 98, 114, 111, 119, 110, 32, 102, 111, 120, 32, 106, 117, 109, 112, 115, 32, 111, 118, 101, 114, 32, 116, 104, 101, 32, 108, 97, 122, 121, 32, 100, 111, 103, 46]`',
io-uring-test/src/tests/register_buf_ring.rs:611:9
```

### DIFF
--- a/io-uring-test/src/tests/register_buf_ring.rs
+++ b/io-uring-test/src/tests/register_buf_ring.rs
@@ -545,12 +545,13 @@ fn buf_ring_read<S, C>(
     ring: &mut IoUring<S, C>,
     buf_ring: &FixedSizeBufRing,
     fd: types::Fd,
+    len: u32,
 ) -> io::Result<GBuf>
 where
     S: squeue::EntryMarker,
     C: cqueue::EntryMarker,
 {
-    let read_e = opcode::Read::new(fd, std::ptr::null_mut(), 0)
+    let read_e = opcode::Read::new(fd, std::ptr::null_mut(), len)
         .offset(0)
         .buf_group(buf_ring.rc.bgid());
 
@@ -580,6 +581,7 @@ where
     }
 
     let result = result as u32;
+    assert_eq!(result, len);
     let flags = cqes[0].flags();
     let buf = buf_ring.rc.get_buf(buf_ring.clone(), result, flags)?;
 
@@ -602,6 +604,7 @@ fn buf_ring_play<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     // by the uring interface.
 
     let text = b"The quick brown fox jumps over the lazy dog.";
+    let len = text.len() as u32;
 
     let normal_check = |buf: &GBuf, bid: Bid| {
         // Verify the buffer id that was returned to us.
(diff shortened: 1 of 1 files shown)


lucab:
It looks like `test_register_buf_ring` was added in https://github.com/tokio-rs/io-uring/pull/151, but normally it does not run in CI due to the old kernel (5.15, while this feature requires 5.19). 
I don't have a baseline to check, but without the length argument this test fails for me on a 5.19 kernel (Ubuntu 22.10 - `5.19.0-29-generic`). The read completes successfully but indeed it reads 0 bytes, so the assertions later on fail.

@quininer @FrankReh @Noah-Kennedy could you please run `test_register_buf_ring` on your machines with and without this patch, and cross-check the behaviors?

FrankReh:
I'm having a look.

FrankReh (approved):
Code changes look good and they work on my 6.1 kernel. They don't seem to introduce any further clippy warnings than are already there.

FrankReh:
@lucab Let me know what assertion was failing for your 5.19 case. Then I can make the commit message a little cleaner because to say it was failing on a 5.19 but not say we haven't seen it fail on 6.0+ but also not say what assertion was failing could cause more confusion to git log readers in the future than is necessary. It is just changing a test so it might not be important but sometimes folks use tests as a starting point for their code along with commit logs. These discussions on the GitHub PR don't make it into the git repo.

io-uring-test/src/tests/register_buf_ring.rs
```diff
@@ -580,6 +581,7 @@ where
     }
 
     let result = result as u32;
+    assert_eq!(result, len);
```
lucab, line 584: When passing `0` as the length for the `opcode::Read` operation, this assertion fails with:
```
test register_buf_ring
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `44`', io-uring-test/src/tests/register_buf_ring.rs:584:5
```

FrankReh: Isn't that a line you added?

lucab:
@FrankReh thanks for the quick feedback! I did add the original assertion failure to the commit message.
That one alone though isn't really direct to decipher. A simpler way to observe the root-cause is through this other assertion that I noted here: https://github.com/tokio-rs/io-uring/pull/173#discussion_r1092347770

io-uring-test/src/tests/register_buf_ring.rs
lucab: Yes, I added this assertion after a bit of drilling down on the original test failure (that one is part of the commit message now). This directly showed me that the problem was at the kernel boundary, as the read operation successfully completed with 0 bytes read.

FrankReh:
@lucab Thanks for that info. I will commit this later today unless one of the other maintainers beats me to it. What you found is super interesting. From the little playing around I just did, it appears something is wrong with 6.0 and 6.1 kernels when that value of 0 is passed in. Maybe it's be design and the actual length of the buffer is passed by some other parameter? 

I'll commit when I have a better feeling why it was passing for the 6.0 and 6.1 kernels. Your fix for all is very reasonable though. Thanks!

lucab:
Looking at kernel history, I encountered this commit which looks relevant:
https://github.com/torvalds/linux/commit/b8c015598c8ef9195b8a2a5089e275c4f64ca999

I don't have a good setup at hand for doing full kernel bisects, but that commit landed in 6.0 which matches the datapoints we collected so far.

FrankReh:
> Looking at kernel history, I encountered this commit which looks relevant:

That is certainly relevant. That explains why the io_uring driver knew the 44 bytes were safe to write into the buffer provided .. because the test was for a buf ring, where the buffer length was provided by the ring api, not by the read operation.

Nice find!

PR #174, opened 2023-02-01
lib: add missing helpers for feature probing

lucab:
This re-syncs the set of helpers that can be used to probe features supported by a ring/kernel, adding support for the following new features:
 * `IORING_FEAT_RSRC_TAGS`
 * `IORING_FEAT_CQE_SKIP`
 * `IORING_FEAT_LINKED_FILE`

All features supported by the latest kernel (6.1) can now be probed through helpers.


quininer:
Thank you!

PR #175, opened 2023-02-02
opcode: add support for IORING_MSG_SEND_FD

lucab:
This introduces a new `MsgRingSendFd` operation in order to wrap the combination of `IORING_OP_MSG_RING` plus `IORING_MSG_SEND_FD` (available since 6.0).

### DIFF
--- a/io-uring-test/src/tests/queue.rs
+++ b/io-uring-test/src/tests/queue.rs
@@ -198,3 +198,110 @@ pub fn test_msg_ring_data<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
 
     Ok(())
 }
+
+pub fn test_msg_ring_send_fd<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> anyhow::Result<()> {
+    use std::os::unix::io::AsRawFd;
+
+    // This test requires both `IORING_OP_MSG_RING` opcode and `IORING_MSG_SEND_FD` flag.
+    // We directly probe for the opcode, but the flag only appeared later
+    // (in kernel 6.0) and cannot be probed at runtime.
+    // Thus, we additionaly check for the `IORING_OP_SEND_ZC` opcode (>= 6.0)
+    // as a proxy instead.
+    require!(
+        test;
+        test.probe.is_supported(opcode::MsgRingData::CODE);
+        test.probe.is_supported(opcode::SendZc::CODE);
+    );
+
+    println!("test msg_ring_send_fd");
+
+    // Ensure there are no fixed files on the test ring, so that slots
+    // allocation will start at 0. Then create a tempfile and register it
+    // (at slot 0).
+    {
+        let _ = ring.submitter().unregister_files();
+        let tmp1 = tempfile::tempfile()?;
+        ring.submitter()
+            .register_files(&[tmp1.as_raw_fd()])
+            .unwrap();
+    }
+
+    // Create a new destination ring.
+    // FD sending only works if the receiving ring already has a slot
+    // registered for it. Thus we register some sparse files (slots 0 and 1).
+    let mut temp_ring = IoUring::new(8)?;
+    temp_ring.submitter().register_files_sparse(2).unwrap();
+
+    // Send the fixed file from the source ring (on slot 0)
+    // to the temp ring (to slot 1).
+    // This will generate two completion events, one on each ring.
+    unsafe {
+        let fd = types::Fd(temp_ring.as_raw_fd());
+        let dest_slot = types::DestinationSlot::try_from_slot_target(1).unwrap();
+        ring.submission()
+            .push(
+                &opcode::MsgRingSendFd::new(fd, types::Fixed(0), dest_slot, 11, 22)
+                    .build()
+                    .into(),
+            )
+            .expect("queue is full");
+    }
+    ring.submit_and_wait(1)?;
+    {
+        let source_cqes: Vec<cqueue::Entry> = ring.completion().map(Into::into).collect();
+        assert_eq!(source_cqes.len(), 1);
+        assert_eq!(source_cqes[0].user_data(), 0);
+        assert_eq!(source_cqes[0].result(), 0);
+        assert_eq!(source_cqes[0].flags(), 0);
+
+        let dest_cqes: Vec<cqueue::Entry> = temp_ring.completion().map(Into::into).collect();
+        assert_eq!(dest_cqes.len(), 1);
+        assert_eq!(dest_cqes[0].user_data(), 22);
+        assert_eq!(dest_cqes[0].result(), 11);
+        assert_eq!(dest_cqes[0].flags(), 0);
+    }
+
+    // Unregister the fixed files from the source ring, then reserve some empty slots
+    // (0 through 3).
+    ring.submitter().unregister_files().unwrap();
+    ring.submitter().register_files_sparse(4).unwrap();
+
+    // Send back the fixed file from the temp ring (on slot 1)
+    // to the temp ring (to slot 2).
+    // This will again generate two completion events, one on each ring.
+    unsafe {
+        let fd = types::Fd(ring.as_raw_fd());
+        let dest_slot = types::DestinationSlot::try_from_slot_target(2).unwrap();
+        temp_ring
+            .submission()
+            .push(&opcode::MsgRingSendFd::new(fd, types::Fixed(1), dest_slot, 33, 44).build())
+            .expect("queue is full");
+    }
+    temp_ring.submit_and_wait(1)?;
+    {
+        let source_cqes: Vec<cqueue::Entry> = temp_ring.completion().map(Into::into).collect();
+        assert_eq!(source_cqes.len(), 1);
+        assert_eq!(source_cqes[0].user_data(), 0);
+        assert_eq!(source_cqes[0].result(), 0);
+        assert_eq!(source_cqes[0].flags(), 0);
+
+        let dest_cqes: Vec<cqueue::Entry> = ring.completion().map(Into::into).collect();
+        assert_eq!(dest_cqes.len(), 1);
+        assert_eq!(dest_cqes[0].user_data(), 44);
+        assert_eq!(dest_cqes[0].result(), 33);
+        assert_eq!(dest_cqes[0].flags(), 0);
+    }
+
+    // Unregister the fixed files from both rings, then repeat again to
+    // ensure that all previous files have been unregistered (on empty rings
+    // the operation will fail with `-ENXIO`).
+    ring.submitter().unregister_files().unwrap();
+    temp_ring.submitter().unregister_files().unwrap();
+    ring.submitter().unregister_files().unwrap_err();
+    temp_ring.submitter().unregister_files().unwrap_err();
+
+    Ok(())
+}
--- a/src/types.rs
+++ b/src/types.rs
@@ -42,6 +42,7 @@ pub(crate) mod sealed {
 
 use crate::sys;
 use bitflags::bitflags;
+use std::num::NonZeroU32;
 use std::os::unix::io::RawFd;
 
 use std::marker::PhantomData;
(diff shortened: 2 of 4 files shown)


io-uring-test/src/tests/queue.rs
```diff
@@ -198,3 +198,95 @@ pub fn test_msg_ring_data<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
...
+    // Looking at the `temp_ring` fdinfo, it seems that the FD is correctly received but
+    // wrongly installed to slot 0. So the rings status at this point is confusing:
+    // ```
+    // [/proc/<PID>/fdinfo/<source ring>]
+    // UserFiles:      1
+    //     0: .tmpy3CuuJ
+    // [/proc/<PID>/fdinfo/<dest ring>]
+    // UserFiles:      1
+    //     0: .tmpy3CuuJ
+    // ```
```
lucab, line 285: @axboe I'm confused by what I'm observing here on the receiving ring. Am I somehow misusing the kernel API?

FrankReh:
@lucab Where did you find documentation describing the sqe fields to use for fixed_slot_src and fixed_slot_dest for the IORING_OP_MSG_RING:IORING_MSMG_SEND_FD operation? I don't even see liburing tests for this feature.

FrankReh:
@lucab Still taking shots in the dark.

Maybe the test has a problem that the dest ring only has a fixed file list of length 1, so can only be indexed with [0] but the test code seems to be trying to send the fd from the source slot [0] to the dest slot [1]. Maybe there is no [1] and there is no error code to tell the operation failed?

lucab:
@FrankReh the interface is quite new and mostly described in the text of commit https://github.com/torvalds/linux/commit/e6130eba8a848a7a6ba6c534bd8f6d60749ae1a9.

Yes there are several implicit/undocumented checks inside the kernel.
To be honest I'm not even sure I'm supposed to be allowed to send to `slot[1]`, however the operation currently succeeds (the kernel seems to internally do a `dest_slot--`). I've tried several other approaches, and the kernel does report proper errnos when things go wrong.
I've pinged Jens to try to get some clarifications on what's the expected usage of this API.

FrankReh:
@lucab I'm not aware that Jens has looked at this Rust code before.

Why do you think the device is subtracting one from the dest slot value? Is it reserving zero to mean something like pick an empty slot or to safeguard the command against accidental misuse? I'm starting to look through the commit from last June - I don't see where the minus one is happening, so maybe a different commit. If you're right and the dest slot is supposed to be one more than actually meant, the MsgRingSendFd struct/function comment should call that out. And the description of types:Fixed is a file descriptor that has been registered with io_uring. But the function you added is using that type for fixed_slot_dest which is just a slot being targeted.

So back to the problem you are seeing: what is wrong with the unix proc output you posted in the comment? Does it show that the fd was in fact copied over to slot 0 successfully? The document doesn't say the old fd should be dropped. Maybe show what the proc output was before the send?

lucab:
To make some details clearer, I initially tried to implement a longer test (some kind of ping-pong between the rings) but eventually had to limit myself to the "one simple send" here as the rest wasn't really working as planned. The "stop testing here" refers to that: we can expand the test later once there is less fog around the expected usage of this API.

I did actually want to send the FD to slot[1] of the destination ring.
slot[0] is already occupied by another different tempfile, and sending to any other slot >=2 fails with `-EINVAL`.
The fact that sending succeeds but there is nothing at slot[1] surprised me. It took a bit of debugging to figure out there is a `slot--` happening under the hood (in [__io_fixed_fd_install](https://github.com/torvalds/linux/blob/v6.1/io_uring/filetable.c#L119) as you mentioned).

For reference, this is the ring status right after registering the initial files, but before any send operations:
```
[/proc/<PID>/fdinfo/<source ring>]
UserFiles:      1
    0: .tmpnjm4zV

[/proc/<PID>/fdinfo/<dest ring>]
UserFiles:      1
    0: .tmpZYqgw8
```

And this is the status after sending completes and the CQEs are delivered:
```
[/proc/<PID>/fdinfo/<source ring>]
UserFiles:      1
    0: .tmpnjm4zV

[/proc/<PID>/fdinfo/<dest ring>]
UserFiles:      1
    0: .tmpnjm4zV
```

lucab:
I plan to play with `IORING_REGISTER_FILE_ALLOC_RANGE` for testing this, but for the moment I ran out of time for that.

lucab:
Uhm, I guess this ends up following the same code flow as other operations on fixed files, e.g. `IORING_OP_ACCEPT`:
```
If the file_index field is set to a positive number, the
file won't be installed into the normal file table as 
usual but will be placed into the fixed file table at 
index file_index - 1.  In this case, instead of returning
a file descriptor, the result will contain either 0 on
success or an error. If the index points to a valid empty 
slot, the installation is guaranteed to not fail. If there
is already a file in the slot, it will be replaced,
similar to IORING_OP_FILES_UPDATE.
```

Even then though, I'm still not sure what is the expected behavior for `fixed_slot_dest=0` and for any send-fd targeting an empty slot.

axboe:
I think the confusion is that sqe->file_index is target_fd + 1, this is done for a number of commands so that file_index == 0 means no direct. Also see ```__io_uring_set_target_fixed_file()``` for the internal helper for that.

I improved the test case a bit, and added some tests for different values for source and target ring:

https://git.kernel.dk/cgit/liburing/commit/?id=c3880f82e2f9640e73fb1936c817cc5980329497

We really need a liburing helper for this use case however, so it's not so hard to grok. I'll add that.

FrankReh:
@lucab So it seems all understood. I'll add some code comments then.

@axboe Welcome, and thanks for stepping in. I didn't find a man page for this feature either. Should we assume the fd is being *copied*, not transferred. As the results above shows the fd ended up in both the original table and the new table. And safe to assume the dest slot value of IORING_FILE_INDEX_ALLOC will have an empty slot picked if possible, and the chosen dest slot is returned to the source ring's cqe result?

axboe:
See last two commits here:

https://git.kernel.dk/cgit/liburing/

one adds a helper, the other adopts it in the test case. There is no man page yet, I'll make sure there is one for when 2.4 is released. We've got some man page additions coming, there should be nothing undocumented after those go in.

FrankReh:
@lucab The arg type for the dest slot could be an enum, having two invariants like AUTO, and Index(u32). In the former case, the value IORING_FILE_INDEX_ALLOC is passed, in latter case, the function adds 1.

axboe:
(Unsubscribing here, pull me in again if needed)

lucab:
Ah, thanks for pointing me towards `test/fd-pass.c` and for expanding it.

So my new understanding after reading that is that 1) userland should pass `dest=target+1` to the kernel 2) `dest=0` is not a valid argument 3) this opcode can only be used to replace slot that have been previously reserved/occupied.

axboe:
Not sure where you get 3) from, you can use it to install a new descriptor into an empty space. But the target ring must have setup a direct descriptor space already, maybe that's what you mean? If it hasn't, it cannot receive descriptors. See the test case, and man page coming obviously.

FrankReh:
@lucab I would agree with that understanding. 3) just means the dest file table works normally. The space needed to be registered first. The example you have registers a single slot, slot 0. Doesn't matter if there was a file desc there already. io_uring is happy to replace one fd with another in where file table slots are concerned. Your example could have used -1 for that slot I think, or there is another register command for registering empty slots.

FrankReh:
I am reminded we don't have any of the helper's that the liburing library calls *direct* helpers:

- accept_direct
- multishot_accept_direct
- openat_direct
- openat2_direct
- close_direct
- socket_direct (with specific and auto flavors)

FrankReh (changes requested):

src/types.rs
```diff
@@ -261,3 +262,59 @@ impl BufRingEntry {
...
+    /// Try to use a given target slot.
+    ///
+    /// Valid slots are in the range from `0` to `u32::MAX - 2` inclusive.
+    pub fn try_from_slot_target(target: u32) -> io::Result<Self> {
+        const MAX_INDEX: u32 = DestinationSlot::AUTO_ALLOC - 2;
+        match target {
+            v if v <= MAX_INDEX => Ok(Self {
+                dest: KernelIndexDest::Index(v + 1),
+            }),
+            v => Err(io::Error::new(
```
quininer, line 300: return `Result<Self, u32> is enough, I prefer to have `io::Error` be treated as a system error

src/types.rs
```diff
@@ -261,3 +262,59 @@ impl BufRingEntry {
...
+
+/// A destination slot for sending fixed resources
+/// (e.g. [`opcode::MsgRingSendFd`](crate::opcode::MsgRingSendFd)).
+#[derive(Debug, Clone, Copy)]
+pub struct DestinationSlot {
+    dest: KernelIndexDest,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum KernelIndexDest {
```
quininer, line 274: This enum seems unnecessary, keeping u32 is fine.

src/types.rs
```diff
@@ -261,3 +262,59 @@ impl BufRingEntry {
...
+                format!("Invalid slot target {}", v),
+            )),
+        }
+    }
+
+    pub(crate) fn kernel_index_arg(&self) -> u32 {
+        use KernelIndexDest::*;
+        match self.dest {
+            Index(v) if v == (Self::AUTO_ALLOC - 1) => {
+                unreachable!("IORING_FILE_INDEX_ALLOC - 1")
```
quininer, line 311: This can't actually happen, but writing a `panic!` here won't make the compiler optimize it away.

It would be nice to remove it or use `debug_assert!`.

quininer (changes requested):

io-uring-test/src/tests/queue.rs
```diff
@@ -198,3 +198,117 @@ pub fn test_msg_ring_data<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
...
+    // (in kernel 6.0) and cannot be probed at runtime.
+    // Thus, we additionaly check for the `IORING_OP_SEND_ZC` opcode (>= 6.0)
+    // as a proxy instead.
+    require!(
+        test;
+        test.probe.is_supported(opcode::MsgRingData::CODE);
+    );
+    require!(
+        test;
+        test.probe.is_supported(opcode::SendZc::CODE);
```
quininer, line 219: You should be able to merge it to 215L.

src/types.rs
FrankReh: I would have suggested SlotSelection at this point.

But you need @quininer's vote so you can ignore me and use u32. I will stop giving opinions here.

lucab: I'm appreciate the reviews from both of you so far :smile: 

I can drop the enum and directly keep the numerical representation. Would you maybe want a `NonZeroU32` here? (the `0` is not possible due to +1 shift enforced by the constructor)

src/types.rs
lucab: I feel that it gets harder for consumers to decipher, but indeed I can switch to that. What is the `u32` in the `Err` variant? The input argument reflected back? If so, would you prefer a `Result<Self, ()>` maybe?

src/types.rs
quininer: `NonZeroU32` sounds good.

src/types.rs
quininer: I think it is common to return to input itself when creation fails. returnin `Option` or `Result<_, ()>` is also good. Or define a simple struct `InvalidInput`.

lucab:
Updated, this should be ready for another review pass.

> I am reminded we don't have any of the helpers that the liburing library calls direct helpers

Do you want to have equivalent Rust wrappers? (I mean in general in the library, outside of this PR)
If so, mind opening a ticket about what those should look like from an high-level signature point-of-view?

FrankReh:
> mind opening a ticket

- #178

src/types.rs
```diff
@@ -261,3 +262,46 @@ impl BufRingEntry {
...
+    pub fn try_from_slot_target(target: u32) -> Result<Self, u32> {
+        const MAX_INDEX: u32 = DestinationSlot::AUTO_ALLOC - 2;
+
+        if target > MAX_INDEX {
+            return Err(target);
+        }
+
+        let kernel_index = target.saturating_add(1);
+        // SAFETY: by construction, always clamped between 1 and IORING_FILE_INDEX_ALLOC-1.
+        let dest = unsafe {
```
quininer, line 296: I think we can avoid unsafe here, it will not cause unnecessary overhead, because it will be optimized away by compiler.

src/types.rs
```diff
@@ -261,3 +262,46 @@ impl BufRingEntry {
...
+        // SAFETY: kernel constant, always non-zero.
+        let dest = unsafe { NonZeroU32::new_unchecked(DestinationSlot::AUTO_ALLOC) };
+        Self { dest }
+    }
+
+    /// Try to use a given target slot.
+    ///
+    /// Valid slots are in the range from `0` to `u32::MAX - 2` inclusive.
+    pub fn try_from_slot_target(target: u32) -> Result<Self, u32> {
+        const MAX_INDEX: u32 = DestinationSlot::AUTO_ALLOC - 2;
```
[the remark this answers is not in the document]
quininer: `panic!` doesn't work in 1.48, we need to write

```rust
#[allow(unconditional_panic)]
const fn unwrap(t: Option<u32>) -> u32 {
    match t {
        Some(t) => t,
        None => [][1]
    }
}
```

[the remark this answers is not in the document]
lucab: Yes, I originally had those kind of const-assertions but I had to remove them once I realized the MSRV here was very old and missing many useful features.
I've now implemented your suggestions, PTAL.

quininer (approved):
LGTM

quininer:
Thank you!

PR #179, opened 2023-02-06
opcode: add support for IORING_OP_SOCKET

lucab:
This adds an ergonomic wrapper for the `IORING_OP_SOCKET` opcode,
which has been introduced in kernel 5.19.

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -116,6 +116,7 @@ fn test<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     tests::net::test_tcp_buffer_select_recvmsg(&mut ring, &test)?;
     #[cfg(not(feature = "ci"))]
     tests::net::test_tcp_buffer_select_readv(&mut ring, &test)?;
+    tests::net::test_socket(&mut ring, &test)?;
 
     // queue
     tests::poll::test_eventfd_poll(&mut ring, &test)?;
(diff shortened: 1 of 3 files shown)


lucab:
@quininer @FrankReh from my side this is ok, but I wasn't sure about the fixed-file angle, so I left it as a draft for now.
Let me know if this is fine as is, or if you prefer to split that part out to a separate PR for further discussion.

quininer:
This looks fine to me, but `destination_slot` needs some documentation explaining it.

FrankReh:
@quininer @lucab Is there a name other than destination_slot that might better convey the fixed table slot nature of the type? The first time it was used, the parameter names made sense to be src and dest, but other operations that take a slot index aren't necessarily thought of as a destination.

lucab:
liburing in `io_uring_prep_socket_direct` calls this parameter `file_index`. I'm not sure it is better at conveying the semantics, but at least there is some prior art to cross-reference.
I've renamed the argument that way and added some docstrings for it. I'm still open to better name suggestions if you have any. Other than that, this PR is now ready for review.

quininer:
thank you!

FrankReh:
Thanks to you both!

FrankReh:
@quininer One question, before a new release is issued, would you accept a new name for the type that is currently named `types::DestinationSlot`?

It could be named `types::FileIndex` with no loss in meaning and might be less confusing for users.

PR #180, opened 2023-02-08
cqueue: add docstrings for multishot and dynamic-buffer flags

lucab:
This adds some docstrings for `cqueue::buffer_select()` and `cqueue::more()`, in order to make them easier to discover (as their usage is basically required for multishot operations).

### DIFF
--- a/src/cqueue.rs
+++ b/src/cqueue.rs
@@ -302,6 +302,11 @@ impl Debug for Entry32 {
     }
 }
 
+/// Return which dynamic buffer was used by this operation.
+///
+/// This corresponds to the `IORING_CQE_F_BUFFER` flag (and related bit-shifting),
+/// and it signals to the consumer which provided contains the result of this
+/// operation.
 pub fn buffer_select(flags: u32) -> Option<u16> {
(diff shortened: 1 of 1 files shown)


quininer:
Thank you!

lucab:
Out of curiosity, is there any reason why those are not implemented directly on `Entry` (and `Entry32`)?
I initially did not find these, and started writing my own version before noticing these. Having them as methods on the structs would make them easier to discover (e.g. via IDE auto-completion).

quininer:
No specific reason, but I'm not sure that being an Entry method is correct semantics.

lucab:
Ack, thanks for the feedback, I won't touch them then.

Just for future reference, my rationale here was that the input of those two is always a value coming from the kernel through `cqe->flags`. So effectively they will be always used in the following way:
https://github.com/tokio-rs/io-uring/blob/dcf6a124926cfb32bc94ff0b617bba6844bf54b6/io-uring-test/src/tests/net.rs#L160
https://github.com/tokio-rs/io-uring/blob/dcf6a124926cfb32bc94ff0b617bba6844bf54b6/io-uring-test/src/tests/net.rs#L511

Issue #178, opened 2023-02-03
fixed file helpers


FrankReh:
This crate doesn't call them helpers but the liburing does and for this, that seems reasonable too.

`io-uring` is missing helper support for direct files in its operations that can accept them.

- [x] openat_direct (being worked by #181)
- [x] openat2_direct (also being worked by #181)
- [x] close_direct (also being worked by #181)
- [x] socket_direct (edit: handled by opcode::Socket with builder method file_index() #179)
- [x] accept_direct (being worked by #184)
- [x] multishot_accept_direct (being worked by #186)

What kind of naming convention would be acceptable for the new class of helper and can they take an enum that represents either a fixed slot number to use or the AUTO choice of the device itself?

lucab:
>  * socket_direct

I started having a look at implementing `IORING_OP_SOCKET`.
I'm slightly puzzled by the `_direct` helpers in liburing, because some of them see to concern **input** FDs (e.g. `close`), while some of them concern **output** FDs (e.g. `openat`).

Maybe the `socket` opcode is an outlier here as it is only concerned with output FDs.
I'm not sure it even needs a dedicated direct helper. Instead a unified helper can be modeled this way:
```rust
opcode!(
    pub struct Socket {
        domain: { i32 },
        socket_type: { i32 },
        protocol: { i32 },
        ;;
        destination_slot: Option<types::DestinationSlot> = None,
    })
```

FrankReh:
@lucab Nice. I agree the signature you are proposing looks like it obviates the need for additional opcodes.

The liburing close_direct is an odd-ball. Yes, it is an input argument and in other cases they use a single bit to indicate the input fd is a fixed slot. But the command is actually doing two things in the device, it is calling close on the file *and* it is unregistering the descriptor from the fixed table.

FrankReh:
Thanks @lucab. Once you got support in for `opcode::Socket` the rest were easy to make follow suite.

lucab:
@FrankReh thank you for your patches!

PR #181, opened 2023-02-10
opcode: add "fixed file" support to openat, openat2 and close

FrankReh:
Following the model for the file_index builder method set in #179.

Addresses three of the opcodes highlighted in #178.

### DIFF
--- a/io-uring-test/src/tests/fs.rs
+++ b/io-uring-test/src/tests/fs.rs
@@ -291,9 +291,7 @@ pub fn test_file_openat2_close_file_index<S: squeue::EntryMarker, C: cqueue::Ent
 
     // Drop two.
     for round in 0..2 {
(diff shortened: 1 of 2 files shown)


FrankReh:
Still a wip while unit tests are missing.

FrankReh:
@quininer I think @lucab is correct in his comment in
- #182 

Do you agree the opcode::Close is better taking a Fixed FD 1st parameter, an no optional file_index method? Would that make it a breaking semver change though?

quininer:
Yes, we should use `UseFixed`, I don't think it's a breaking change. I intentionally designed these two sealed traits so that `impl UseFd` can be changed to `impl UseFixed` at any time.

quininer (changes requested):
about close opcode

quininer (approved):
LGTM

PR #186, opened 2023-02-11
opcode: add support for multi-shot accept

FrankReh:


FrankReh:
Leaving as a draft until I have a unit test for it.

FrankReh:
@lucab @quininer Curious what you two think about just using an optional bool as shown in this PR and what you'd like the name of it to be, if not `fixed`. Other possible names I can think of are `direct`, `registered` or `fixed_table`.

When the accept is a multi-shot accept, the io_uring device will create regular file descriptors if the file_index is left zero or will create fixed file table entries if the file_index is set to IORING_FILE_INDEX_ALLOC. Actual slot indexes are not allowed as the operation is designed to return multiple CQE for multiple accepted connections, one slot index would be nonsensical.

lucab:
I don't have much experience with `IORING_OP_ACCEPT`, but that said I don't see major drawback in the way you modeled it here. The boolean flag looks fine to me.

With regards to naming, you can perhaps keep aligned with the underlying flag and make it a `allocate_file_index: bool`.

src/opcode.rs
```diff
@@ -583,6 +583,36 @@ opcode!(
...
+
+    pub fn build(self) -> Entry {
+        let AcceptMulti { fd, fixed, flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.ioprio = sys::IORING_ACCEPT_MULTISHOT as u16;
+        // No out SockAddr is passed for the multishot accept case.
+        // The user should perform a syscall to get any resulting connection's remote address.
```
lucab, line 607: Out of curiosity, how would this be performed when using the fixed file allocation?

FrankReh: Not sure I understand the question. What do you mean *performed*? If you're asking about the kernel side, it means the user can have preregistered a fixed file table, say of 5000 slots. Then the multishot accept op can accept up to 5000 connections. If the user's app, or library closes some of those in time, then even more as the slots will get reused by the device's selection process. If you're asking about the app or library side, it means the only operations suitable for those fixed file table file descriptors are the ones the io_uring device itself supports. The user won't be able to get a real file descriptor from them so won't be able to issue other syscalls on them.

FrankReh:
> With regards to naming, you can perhaps keep aligned with the underlying flag and make it a `allocate_file_index: bool`.

I like your suggestion better than `fixed`. Am making that change.

src/opcode.rs
FrankReh: Perhaps you meant, how would `tokio-uring` handle the multi responses? The API will stream the connections.

lucab: Oh sorry, I should have been more explicit with that question.
What I meant was, what is the actual mechanism to "to get any resulting connection's remote address" when using fixed file allocation? My understanding is "perform a syscall" likely refers to `getpeername(2)` here when using FDs, but I'm not seeing an equivalent io-uring operation for fixed files.
(It isn't really a question on this specific PR, I know, but I feel like I'm maybe missing a piece here)

FrankReh: Yea, you are write. So the fixed descriptor forms of accept and multishot accept may only be useful a small percentage of the time - when other syscalls aren't necessary. When one wants to know the peer, or set other things about the socket, the user will have used the regular file descriptor form of accept and multishot accept. The user has the ability then to pick a location manually and use the register/file-update mechanism and still get the benefits of skipping the reference count increment/decrement that otherwise has to be done for each read and write.

Guess I'll work on that.

quininer:
this looks fine to me

quininer (approved):

Issue #187, opened 2023-02-11
missing multishot commands


FrankReh:
I'm going to be adding more multishot commands, some or all in separate PRs. But I wanted a place where we could discuss the naming of these opcodes and options. I'm going to list the names that come from the liburing API as a starting point.

- [x] recv_multishot(sockfd, buf_group, flags) (see #188)
- [x] recvmsg_multishot (see #192)
- [x] poll_multishot(fd, flags) (see #191)

The multishot version of recv should not take a buffer pointer or length so a separate opcode seems warranted to me, maybe `RecvMulti` as a convention I think we're using is to place modifiers in a name after the primary part of the name? If we wanted to
use the same opcode, we could add an optional buf_group field, but a buf_group id value of 0 is allowed so we might have to wrap that optional type in an Option in that case or default it to a max value.

The multishot version of recvmsg I'm still confused about. I think parts of the msghdr are still relevant to the function even though the buffer it is designed to point to would not be. I would be tempted to tackle this one last because it's not clear to me from even the liburing documentation how it works or what it does. It might require looking through the kernel source or asking on the liburing repo.

The poll_multishot ends up taking the same arguments as the single-shot poll_add operation, so maybe that doesn't deserve its own opcode here, maybe an  optional `multishot` boolean? But if you want a separate opcode, how about the name `PollMulti` to go with the `PollAdd` that is there now?

@quininer @lucab

lucab:
> The multishot version of recvmsg I'm still confused about. I think parts of the msghdr are still relevant to the function even though the buffer it is designed to point to would not be. I would be tempted to tackle this one last because it's not clear to me from even the liburing documentation how it works or what it does. It might require looking through the kernel source or asking on the liburing repo.

I was playing with this one and I have a locally working branch (on kernel 6.1), but I still need to polish and write tests before submitting.

What I've found out so far is that:
 * the userland needs to pass some real `msghdr`, a NULL pointer is invalid and rejected by kernel
 * the structure is not used for storing results from the kernel, but only read from for input arguments
 * the kernel only uses two values from that structure (name length and control length), this way:
   - `msg_namelen` is the reserved fixed length (in bytes) to store the name result, in the provided-buffer (0 is a valid input)
   - `msg_namelen` is the reserved fixed length (in bytes) to store the control data, in the provided-buffer (0 is a valid input)
[...]

PR #189, opened 2023-02-12
types: export multishot header(io_uring_recvmsg_out) to user.

glenn-wang:


src/types.rs
```diff
@@ -330,3 +330,37 @@ impl DestinationSlot {
         self.dest.get()
     }
 }
+
+/// multishot header struct
+/// multishot recvmsg layout: <header><name><control><payload>
+/// https://lwn.net/Articles/901221/ 
+#[derive(Default, Debug, Clone, Copy)]
+#[repr(transparent)]
+pub struct RecvmsgOut(sys::io_uring_recvmsg_out);
```
lucab, line 339: I was playing with multishot recvmsg (see https://github.com/tokio-rs/io-uring/issues/187#issuecomment-1426993071) and from a consumer point-of-view I have the feeling this is not the most useful data-structure to have.

What I'm currently playing with instead looks like this:
```rust
pub struct RecvMsgOut<'buf> {
    buf: &'buf [u8],
    header: sys::io_uring_recvmsg_out,
    msghdr_name_len: u32,
    msghdr_control_len: u32,
}

impl<'buf> RecvMsgOut<'buf> {
    pub fn parse(buf: &'buf [u8], name_len: u32, control_len: u32) -> Result<Self> { ... }
}
```

In plain words, my current understanding is that to properly handle the result of multishot recvmsg the consumer needs to glue and store together all of the following:
 * the raw data output in the provided buffer
 * the `io_uring_recvmsg_out` content in that data (validated and parsed upfront)
 * the `namelen` and `controllen` that were provided in the SQE `msghdr` (not the whole struct content, just those two are strictly required)

If you are not in a hurry, I can try to finish writing tests and submit what I have next week.

FrankReh: Thank you both. As I just wrote in #187, I look forward to the work in @lucab 's branch, sounds like a lot was figured out already and we will all learn a lot looking at the unit test.

@glenn-wang Thank you for still being interested. I don't know if you know this but this `io-uring` branch is not generally meant for writing a final application but that doesn't mean it can't be done. In fact, I can imagine there are some proprietary solutions already that use this crate with either std threads or their own choice of async runtime.

You can consider putting this in draft mode, we aren't likely to be using the data structure.

I don't know where your past experience lies, but if you have any experience with qemu test runners, I know the maintainer of this repo has asked for help in getting the CI run with a newer version of the kernel. A 6.1 version would be very good, with a way to get it rebuilt as it is now considered an LTS with quite a few years of support planned - so any bug fixes to the io_uring device code will be reflected in rebuilds. I don't think new features will be back ported to it though, so that's a bummer but understandable.

glenn-wang: > I don't know if you know this but this io-uring branch is not generally meant for writing a final application but that doesn't mean it can't be done. In fact, I can imagine there are some proprietary solutions already that use this crate with either std threads or their own choice of async runtime.

I know this. For example , [tokio-rs/tokio-uring](https://github.com/tokio-rs/tokio-uring) is very good. But that project may not encapsulate all the features of io-uring, and it is lagging behind of tokio-rs/io-uring.

> I don't know where your past experience lies, but if you have any experience with qemu test runners, I know the maintainer of this repo has asked for help in getting the CI run with a newer version of the kernel. 

I have no experience of qemu test runners so have no ability to make a help.

FrankReh: Okay. Didn't mean to offend. You are welcome to use this repo for your own purposes of course.

lucab: The work-in-progress logic I was talking about is now up at https://github.com/tokio-rs/io-uring/pull/192.

PR #191, opened 2023-02-12
opcode: add support for IORING_POLL_ADD_MULTI

FrankReh:
Adds a `multi` optional field to the `PollAdd` opcode.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -346,25 +346,31 @@ impl WriteFixed {
 opcode!(
     /// Poll the specified fd.
     ///
-    /// Unlike poll or epoll without `EPOLLONESHOT`, this interface always works in one shot mode.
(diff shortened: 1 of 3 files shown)


src/opcode.rs
```diff
@@ -346,25 +346,31 @@ impl WriteFixed {
...
+    /// Since 5.13, the multi-shot mode is available.
     #[derive(Debug)]
     pub struct PollAdd {
         /// The bits that may be set in `flags` are defined in `<poll.h>`,
         /// and documented in `poll(2)`.
         fd: { impl sealed::UseFixed },
-        flags: { u32 }
+        flags: { u32 },
         ;;
+        multi: bool = false
```
quininer, line 360: I think some docs is needed here

FrankReh: Docs added.

PR #192, opened 2023-02-13
types: add RecvMsgOut helper structure

lucab:
This introduces a `RecvMsgOut` structure which allows consumers to
safely parse the buffered result from a multishot `IORING_OP_RECVMSG`
completion event.

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -1235,3 +1235,147 @@ pub fn test_socket<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
 
     Ok(())
 }
+
+pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> anyhow::Result<()> {
+    // Multishot recvmsg was introduced in 6.0, like `SendZc`.
+    // We cannot probe for the former, so we check for the latter as a proxy instead.
+    require!(
+        test;
+        test.probe.is_supported(opcode::RecvMsg::CODE);
+        test.probe.is_supported(opcode::ProvideBuffers::CODE);
+        test.probe.is_supported(opcode::SendZc::CODE);
+    );
+
+    println!("test udp_recvmsg_multishot");
+
+    let (socket_slot, socket_addr) = {
+        // `:0` means "pick up a random available port number", which should
+        // help avoiding test flakes if a static port is already in use.
+        let server_sock = std::net::UdpSocket::bind("127.0.0.1:0")?;
+        ring.submitter()
+            .register_files(&[server_sock.as_raw_fd()])?;
+        let addr = server_sock.local_addr().unwrap();
+        (io_uring::types::Fixed(0), addr)
+    };
+
+    // Provide 2 buffers in buffer group `33`, at index 0 and 1.
+    // Each one is 512 bytes large.
+    const BUF_GROUP: u16 = 33;
+    const SIZE: usize = 512;
+    let mut buffers = [[0u8; SIZE]; 2];
+    for (index, buf) in buffers.iter_mut().enumerate() {
+        let provide_bufs_e = io_uring::opcode::ProvideBuffers::new(
+            buf.as_mut_ptr(),
+            SIZE as i32,
+            1,
+            BUF_GROUP,
+            index as u16,
+        )
+        .build()
+        .user_data(11)
+        .into();
+        unsafe { ring.submission().push(&provide_bufs_e)? };
+        ring.submitter().submit_and_wait(1)?;
+        let cqes: Vec<io_uring::cqueue::Entry> = ring.completion().map(Into::into).collect();
+        assert_eq!(cqes.len(), 1);
+        assert_eq!(cqes[0].user_data(), 11);
+        assert_eq!(cqes[0].result(), 0);
+        assert_eq!(cqes[0].flags(), 0);
+    }
+
+    // This structure is actually only used for input arguments to the kernel
+    // (and only name length and control length are actually relevant).
+    let mut msghdr: libc::msghdr = unsafe { std::mem::zeroed() };
+    msghdr.msg_namelen = 32;
+    msghdr.msg_controllen = 0;
+
+    // TODO(lucab): make this more ergonomic to use.
+    const IORING_RECV_MULTISHOT: u16 = 2;
+
+    let recvmsg_e = io_uring::opcode::RecvMsg::new(socket_slot, &mut msghdr as *mut _)
+        .ioprio(IORING_RECV_MULTISHOT)
+        .buf_group(BUF_GROUP)
+        .build()
+        .flags(io_uring::squeue::Flags::BUFFER_SELECT)
+        .user_data(77)
+        .into();
+    unsafe { ring.submission().push(&recvmsg_e)? };
+    ring.submitter().submit().unwrap();
+
+    let client_socket: socket2::Socket = std::net::UdpSocket::bind("127.0.0.1:0").unwrap().into();
+    let client_addr = client_socket
+        .local_addr()
+        .unwrap()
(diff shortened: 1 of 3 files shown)


FrankReh:
@lucab This is nice. Thanks for getting to it so quickly. Figuring out the struct being returned and how to test all that isn't a trivial job.

My preference would be for a new opcode: RecvMsgMulti but let @quininer weigh in.

io-uring-test/src/tests/net.rs
```diff
@@ -1093,3 +1093,157 @@ pub fn test_socket<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
...
+
+    let recvmsg_e = io_uring::opcode::RecvMsg::new(socket_slot, &mut msghdr as *mut _)
+        .ioprio(IORING_RECV_MULTISHOT)
+        .buf_group(BUF_GROUP)
+        .build()
+        .flags(io_uring::squeue::Flags::BUFFER_SELECT)
+        .user_data(77)
+        .into();
+    unsafe { ring.submission().push(&recvmsg_e)? };
+    ring.submitter().submit().unwrap();
```
FrankReh, line 1164: You didn't have to submit_and_wait(3) to ensure you got all 3?

FrankReh (approved):
That test is very nice. If you're sure you don't need to submit and wait for 3, it's fine with me.

Well, I do prefer a new opcode were used to convey this different operation to the user as a few things need to all be done correctly but it does work as you've shown.

io-uring-test/src/tests/net.rs
lucab: There should be no CQEs at this point, the kernel does not explicitly ACK the multishot SQE and the completion events only start coming in once the first incoming UDP datagram hits the server socket.

But I think you are generally right, I can add an explicit sync point here and make sure there are no CQEs pending here.

FrankReh: I meant to wait for 3 after the two datagrams were sent, before collecting the cqes. Other tests do that, I think to ensure it waited long enough. Some test beds might change the timing, but maybe you know more than I was willing to assume about how the kernel ip stack handles the loopback messages so that's fine with me. I'm just surprised it worked.

io-uring-test/src/tests/net.rs
```diff
@@ -1093,3 +1093,157 @@ pub fn test_socket<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
...
+    }
+
+    // This structure is actually only used for input arguments to the kernel
+    // (and only name length and control length are actually relevant).
+    let mut msghdr: libc::msghdr = unsafe { std::mem::zeroed() };
+    msghdr.msg_namelen = 32;
+    msghdr.msg_controllen = 0;
+
+    // TODO(lucab): make this more ergonomic to use.
+    const IORING_RECV_MULTISHOT: u16 = 2;
```
FrankReh, line 1154: By creating a new opcode, like I'm currently working in PR 193, I'm also avoiding the need to make a bit like this public and then having to document it and provide an example in the original opcode of how to accomplish the different functionality instead. It's all just code and won't make a different to a higher level library that uses it, once it is in place though. So six of one, half a dozen of the other. If @quininer likes it this way, I'll probably change my approach in 193 to conform.

io-uring-test/src/tests/net.rs
lucab: I see. Yes there is indeed a theoretical race after the socket `send_to()`.
A `submit_and_wait(3)` before collecting CQEs should hopefully be enough to avoid that issue, I'll play with that to check. Thanks for the suggestion and the further explanation!

src/types.rs
```diff
@@ -330,3 +331,119 @@ impl DestinationSlot {
...
+    ///
+    /// When `true`, data returned by `control_data()` is truncated and
+    /// incomplete.
+    pub fn is_control_data_truncated(&self) -> bool {
+        self.header.controllen > self.msghdr_control_len
+    }
+
+    /// Message control data, with the same semantics as `msghdr.msg_control`.
+    pub fn control_data(&self) -> &[u8] {
+        let control_start = Self::DATA_START + self.msghdr_name_len as usize;
```
quininer, line 423: 398L maybe take `header.namelen`, but use `msghdr_name_len` unconditionally here, which is somewhat inconsistent

quininer (changes requested):

src/types.rs
lucab: This should be correct as is, because the fixed length of the **name field** is `msghdr_name_len`, while the dynamic length of the **name data** is the minimum between the field length and the length of the actual result data. In this case we want to skip the whole **field**, whose length is fixed upfront regardless of the result.

But I agree this may be confusing. I've added a docstring on both `msghdr_name_len` and `msghdr_control_len` which hopefully could clarify the semantics.

io-uring-test/src/tests/net.rs
FrankReh: Agreed.

src/types.rs
```diff
@@ -330,3 +331,119 @@ impl DestinationSlot {
...
+    pub fn is_name_data_truncated(&self) -> bool {
+        self.header.namelen > self.msghdr_name_len
+    }
+
+    /// Message control data, with the same semantics as `msghdr.msg_control`.
+    pub fn name_data(&self) -> &[u8] {
+        let name_start = Self::DATA_START;
+        let name_size = u32::min(self.header.namelen, self.msghdr_name_len);
+        let name_end = name_start + name_size as usize;
+        &self.buf[name_start..name_end]
```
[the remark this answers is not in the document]
lucab: Could you add some more details about what you are suggesting here?
My feeling is that what you want to do goes against the limit of having self-references within `RecvMsgOut` (the buffer, plus the subfields pointing within it), but maybe I misunderstood what you have in mind.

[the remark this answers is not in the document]
lucab: On the topic of performances, I did tweak the logic a bit so that all accessor methods can use unchecked/saturating math.

FrankReh (approved):
LGTM. I didn't go through the unit test with a fine-tooth comb, but it looks like before only better.

src/types.rs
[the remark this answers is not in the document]
quininer: It is an immutable reference, there is no self-referencing problem. We also don't need to keep original buf, we can use `split_at` to split.

src/types.rs
```diff
@@ -330,3 +330,134 @@ impl DestinationSlot {
...
+    /// (only `msg_namelen` and `msg_controllen` fields are relevant).
+    pub fn parse(buffer: &'buf [u8], msghdr: &libc::msghdr) -> Result<Self, ()> {
+        if buffer.len() < std::mem::size_of::<sys::io_uring_recvmsg_out>() {
+            return Err(());
+        }
+        // SAFETY: buffer (minimum) length is checked here above.
+        let header: sys::io_uring_recvmsg_out =
+            unsafe { std::ptr::read_unaligned(buffer.as_ptr() as _) };
+
+        let msghdr_name_len = msghdr.msg_namelen as _;
```
quininer, line 367: It's better to use `.into()` whenever possible

src/types.rs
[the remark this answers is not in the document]
lucab: Ack, thanks, initially I didn't understand what you were suggesting. I've changed the parsing logic so that the buffer is split upfront and the structure only keeps references to the three specific data slices.

src/types.rs
lucab: This is a `u32 -> usize` conversion, which is always safe to perform on Linux but unfortunately not yet covered in Rust by `Into` (see https://github.com/rust-lang/rust/issues/70460).

Additionally there are some mismatches across different libc, so that the source fields are `u32` on some libc and `usize` on other libc (so actually not really casting). This is all very unfortunate, but `as _` is pretty much a required hack to appease rustc and clippy in a portable way across libc's.

quininer (approved):
LGTM

FrankReh:
Thanks everyone!

PR #193, opened 2023-02-13
opcode: add optional fixed_buf to SendZc

FrankReh:
The SendZc opcode now takes an optional fixed_buf parameter. Specify it as builder.fixed_buf(Some(id)) as a value of 0 is a valid fixed buffer index.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1604,16 +1604,16 @@ opcode!(
     /// A fixed (pre-mapped) buffer can optionally be used from pre-mapped buffers that have been
     /// previously registered with [`Submitter::register_buffers`](crate::Submitter::register_buffers).
     pub struct SendZc {
-        /// The `buf_index` is an index into an array of fixed buffers,
-        /// and is only valid if fixed buffers were registered.
-        ///
-        /// The buf and len arguments must fall within a region specified by buf_index in the
-        /// previously registered buffer. The buffer need not be aligned with the start of the
-        /// registered buffer.
         fd: { impl sealed::UseFixed },
         buf: { *const u8 },
         len: { u32 },
         ;;
+        /// The `buf_index` is an index into an array of fixed buffers, and is only valid if fixed
+        /// buffers were registered.
+        ///
+        /// The buf and len arguments must fall within a region specified by buf_index in the
+        /// previously registered buffer. The buffer need not be aligned with the start of the
+        /// registered buffer.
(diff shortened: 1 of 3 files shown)


FrankReh:
Addresses
- #185

src/opcode.rs
```diff
@@ -1610,6 +1610,43 @@ opcode!(
...
+        let SendZcFixed { fd, buf, len, buf_index, flags, zc_flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_2.addr = buf as _;
+        sqe.len = len;
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
+        sqe.__bindgen_anon_4.buf_index = buf_index;
+        sqe.ioprio = zc_flags | sys::IORING_RECVSEND_FIXED_BUF as u16;
```
quininer, line 1645: It looks like kernel doesn't like using separate opcodes anymore, maybe we can add an optional method on `SendZc` to follow this style?

FrankReh: Sure. `buf_index` or `fixed_buf_index`?

FrankReh: But a value of 0 is valid buf_index. So not quite sure how to tell. Maybe set it to the max by default? Or allow negative -1
or wrap the type in `Other`?

quininer: I think can use `Option<>`

quininer: buf_index is fine, need some document.

FrankReh: Done.

src/opcode.rs
```diff
@@ -1600,19 +1600,29 @@ opcode!(
 
 opcode!(
     /// Send a zerocopy message on a socket, equivalent to `send(2)`.
+    ///
+    /// A fixed (pre-mapped) buffer can optionally be used from pre-mapped buffers that have been
+    /// previously registered with [`Submitter::register_buffers`](crate::Submitter::register_buffers).
     pub struct SendZc {
+        /// The `buf_index` is an index into an array of fixed buffers,
```
quininer, line 1607: Should the document be placed on `buf_index` field?

FrankReh: Done. I had taken my cue from the ReadFixed example. Should I issue a PR to move that field's comment?

PR #195, opened 2023-02-14
user doc: update sqpoll and other feature user documentation

FrankReh:
The documentation about needing to be root for the setup_sqpoll function had been outdated.

Other documentation for feature flags was either missing or seemed less informative than the info in the newer liburing io_uring_setup(2) man page.


FrankReh:
I took the new documentation straight from the liburing's io_uring_setup(2) man page. I don't know if we want to start referring to such man pages in this crate's documentation.

FrankReh:
Another option would be to link to an established website's man page for io_uring_setup(2). I am partial to https://man.archlinux.org/man/io_uring_setup.2

PR #196, opened 2023-02-14
fix submit_with_args to match submit_and_wait

FrankReh:
The public function `submit_with_args` had fallen behind in logic that I believe should be a duplicate of `submit_and_wait`.


FrankReh:
Also noticed there is no public documentation for `submit_with_args`. I could add that if you like.

FrankReh:
Also noticed there is no equivalent for the newer `io_uring_enter2` system call but I guess no one has missed it yet.

PR #201, opened 2023-02-17
opcode: add RecvMsgMulti

FrankReh:
Adds the RecvMsgMulti opcode.

Updated the unit test to use it.

Also fixes the RecvMsg handling of the seq.flags field. It wouldn't have handled a fixed file descriptor.

Also improves the documentation to RecvMsg, to better describe the buf_group parameter and the *Multi* aspect of the operation.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1152,7 +1152,19 @@ opcode!(
 opcode!(
     /// Receive multiple messages from a socket, equivalent to `recv(2)`.
     ///
+    /// Parameter:
+    ///     buf_group: The id of the provided buffer pool to use for each received message.
+    ///
     /// MSG_WAITALL should not be set in flags.
+    ///
+    /// The multishot version allows the application to issue a single receive request, which
+    /// repeatedly posts a CQE when data is available. Each CQE will take a buffer out of a
+    /// provided buffer pool for receiving. The application should check the flags of each CQE,
+    /// regardless of its result. If a posted CQE does not have the IORING_CQE_F_MORE flag set then
+    /// the multishot receive will be done and the application should issue a new request.
+    ///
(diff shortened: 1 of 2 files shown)


FrankReh:
I used some spacing in the documentation for the parameters. I don't know how that will look when it is actually rendered.

src/opcode.rs
```diff
@@ -509,6 +509,58 @@ opcode!(
...
+    /// of the rest of the buffer in combination with the initial msghdr structure submitted with
+    /// the request. Use [`types::RecvMsgOut`] to parse the data received and access its
+    /// components.
+    ///
+    /// The recvmsg multishot variant is available since kernel 6.0.
+
+    #[derive(Debug)]
+    pub struct RecvMsgMulti {
+        fd: { impl sealed::UseFixed },
+        msg: { *mut libc::msghdr },
```
lucab, line 539: I think this doesn't strictly need to be `mut`.

FrankReh: Done. Thanks.

src/opcode.rs
```diff
@@ -509,6 +509,58 @@ opcode!(
...
+    /// should issue a new request.
+    ///
+    /// Unlike [`RecvMsg`], this multishot recvmsg will prepend a struct which describes the layout
+    /// of the rest of the buffer in combination with the initial msghdr structure submitted with
+    /// the request. Use [`types::RecvMsgOut`] to parse the data received and access its
+    /// components.
+    ///
+    /// The recvmsg multishot variant is available since kernel 6.0.
+
+    #[derive(Debug)]
```
quininer, line 536: This should remove?

quininer: It looks like some opcode is not implemented Debug.

PR #209, opened 2023-03-04
ci: use host kernel / remove cross

rushilmehra:
Until we figure out how to get the latest kernel working in CI, let's use the hosts's kernel since it's newer than what we use currently

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,23 +14,32 @@ jobs:
           - x86_64-unknown-linux-gnu
           - aarch64-unknown-linux-gnu
 
+        os:
+          - name: openbsd
+            architecture: x86-64
+            version: '9.2'
+
+          - name: openbsd
+            architecture: arm64
+            version: '7.2'
+
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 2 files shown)


RushilMehra:
Figuring out how we can support running tests for ARM. I found https://github.com/marketplace/actions/arm-runner which looks interesting

rushilmehra:
@quininer Any idea why the aarch64 test build fails with `Error: Function not implemented (os error 38)`? According to https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md `ubuntu-latest` is running the `5.15.0-1034-azure` kernel, which is higher than the previous CI kernel version of `5.10.0-4` (https://github.com/quininer/cross-newkernel/blob/master/linux-image.sh#L9)

quininer:
I'm not sure, you can first check which system call returned an os error 38.

rushilmehra:
Yeah I figured out the issue. Cross will handle cross compiling a binary given a target arch, and that works great for simply running unit tests, but it won't work for running the `io-uring-test` package since that needs a kernel that supports the target arch. We'll need a VM solution for this to work, so I might as well just come up with a way to test fresher kernels while i'm at it

PR #210, opened 2023-03-08
docs: add docs for shutdown op

RushilMehra:
Tests coming soon in a followup PR

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1375,6 +1375,8 @@ opcode!(
 // === 5.11 ===
 
 opcode!(
+    /// Shut down all or part of a full duplex connection on a socket, equivalent to `shutdown(2)`.
+    /// Available since kernel 5.11.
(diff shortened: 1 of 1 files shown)


src/opcode.rs
```diff
@@ -1375,6 +1375,7 @@ opcode!(
 // === 5.11 ===
 
 opcode!(
+    /// Shut down all or part of a full duplex connection on a socket, equivalent to `shutdown(2)`.
```
quininer, line 1378: It would be good to add available kernel version in document.

RushilMehra: Added, is it worth doing this everywhere? Most docs entries don't have the kernel version explicitly mentioned

quininer: Yes, all API should have version docs, just like manpage.

quininer:
Thank you!

PR #211, opened 2023-03-11
ci: use host kernel for tests

rushilmehra:
Until we find an ergonomic way to virtualize aarch64 within github
actions, let's stick with only running the test package on x86. Unit
tests will run on both architectures via cross.

This bumps the kernel version used for tests from 5.10 to 5.15. To run
newer kernel versions (and to reintroduce running the test package on
aarch64) we'll need to revisit virtualization with QEMU.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,11 +16,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: nightly
-          profile: minimal
-          override: true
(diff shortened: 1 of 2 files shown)


rushilmehra:
@quininer Frankly there aren't any good solutions for virtualizing aarch64 within github actions. Your current solution of using cross + QEMU is probably the best we're gonna get. If we want to keep running the test package on aarch64, I can open a PR to your custom image to bump the kernel version and we can keep using that. In any case, we'll have to do something similar that long term to actually support newer kernels.

quininer:
I think this is good enough for now, thank you.

I will accept PR to update cross custom image.

PR #213, opened 2023-03-13
sys: Update to linux 6.2

quininer:


FrankReh (approved):
LGTM. At some point a new release is warranted. But schedule that with your other plans. Thank you.

quininer:
There are currently no API changes, so there is no need to release a new version. Unless you will be adding an API to expose IORING_SEND_ZC_REPORT_USAGE. :)

PR #216, opened 2023-03-16
Add support for ring mapped buffers

thinkharderdev:
DISCLAIMER: I'm very new to C interop so this may be completely wrong :)

Add support for ring mapped buffer:
1. Add a rust type to wrap managing ring mapped buffers
2. Add example that re-implements the echo server example using ring mapped buffers and multishot receive.


quininer:
This looks similar to [InnerBufRing here](https://github.com/tokio-rs/io-uring/blame/master/io-uring-test/src/tests/register_buf_ring.rs#L92), I have reservations about adding this level of abstraction.
I'm not sure this is generic enough code, as different buffer manager may want different buffer struct.

@FrankReh Do you think this is useful?

FrankReh:
I think it's too opinionated for a crate that projects rely on to provide bindings. The code is already in the test file for anyone who wants to start their own version. And by leaving it in the test, it is not something that needs to be maintained.

I think it belongs at a higher level. I vote on letting someone else maintain and possibly enhance it.

thinkharderdev:
> I think it's too opinionated for a crate that projects rely on to provide bindings. The code is already in the test file for anyone who wants to start their own version. And by leaving it in the test, it is not something that needs to be maintained.
> 
> I think it belongs at a higher level. I vote on letting someone else maintain and possibly enhance it.

Fair enough, one thing that would need to be exposed is `sys::IORING_CQE_BUFFER_SHIFT` in order to implement this outside the crate. Maybe I can rework this PR by exposing that publicly and then moving the impl into the example?

FrankReh:
> Fair enough, one thing that would need to be exposed is `sys::IORING_CQE_BUFFER_SHIFT` in order to implement this outside the crate. It's how I did it in tokio-uring.

cqueue::buffer_select is there for the purpose of extracting the id from the flags. I haven't needed to expose that constant.

This crate hasn't tried to provide examples of other API uses.

If the API is sufficient, the biggest help is in improving documentation around the functions that interested you. If you think something could have been more clear or more helpful.

PR #217, opened 2023-03-16
lib: add `register_buffers_*` family of functions

fathyb:
This introduces:
- `Submitter::register_buffers_tags` corresponding to  `io_uring_register_buffers_tags`
- `Submitter::register_buffers_sparse` corresponding to  `io_uring_register_buffers_sparse`
- `Submitter::register_buffers_update_tag` corresponding to  `io_uring_register_buffers_update_tag`

PR #219, opened 2023-03-18
lib: make `register_buffers` unsafe

fathyb:
This breaking change is required as calling `register_buffers` with incorrect inputs may introduce undefined behaviour.

This PR also changes the method documentation to use `fixed buffers` instead of `user buffers` to match the `io_uring` man pages terminology, related to #217.

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -206,7 +206,7 @@ pub fn test_tcp_zero_copy_send_fixed<S: squeue::EntryMarker, C: cqueue::EntryMar
         iov_len: buf0.len() as _,
     };
     let iovecs = [iovec];
-    ring.submitter().register_buffers(&iovecs).unwrap();
+    unsafe { ring.submitter().register_buffers(&iovecs).unwrap() };
 
     let text_len = text.len();
 
--- a/src/submit.rs
+++ b/src/submit.rs
@@ -172,10 +172,9 @@ impl<'a> Submitter<'a> {
     ///
     /// # Safety
     ///
-    /// This function is unsafe because improper use may lead to memory problems.
-    /// For example, a use-after-free may occur if `iov_base` contains a pointer freed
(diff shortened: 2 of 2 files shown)


FrankReh:
I don't mind the breaking change and you raise a good point for a crate that wants to be sound. Could you provide a Safety comment that was more in the affirmative of what the caller has to do to maintain the safety contract? Saying it should not be used improperly is more vague than other tokio-maintained unsafe comments.

fathyb:
I've tweaked the message to match the one on `Submission::push`, let me know what you think!

FrankReh:
Yes, looks better to me. @quininer is in a different timezone. Have to wait for their review.

quininer:
This makes sense. I will release a 0.6-pre first so that I can merge other break changes in the meantime.

quininer:
I noticed that [`register_buf_ring`](https://github.com/tokio-rs/io-uring/blob/master/src/submit.rs#L418) is still not marked unsafe and `ring_addr` is not an address type. maybe we should change it too. cc @FrankReh

FrankReh:
> I noticed that [`register_buf_ring`](https://github.com/tokio-rs/io-uring/blob/master/src/submit.rs#L418) is still not marked unsafe and `ring_addr` is not an address type. maybe we should change it too. cc @FrankReh

Yes. I can see the value in making those changes now too.

You think the ring_addr type should be made `*const BufRingEntry`? You give me a day or two to make a PR?

quininer:
`*const BufRingEntry` is fine, please feel free to open PR. :)

FrankReh:
> `*const BufRingEntry` is fine, please feel free to open PR. :)

Checkout PR 221. I didn't include a type change because as the test code shows, a user of the library is not always really starting with a BufRingEntry array or vector and casting to a u64 is more directly what the io_uring API calls for.

But I'm on the fence. If you want it, I will do that too. I'm just not sure so wanted you to take a second look first.

PR #221, opened 2023-03-21
lib: make `register_buf_ring` unsafe

FrankReh:
And clean up a clippy warning in the test case while we're at it.


quininer:
I'm not sure what type is best, so I would merge this first.

PR #222, opened 2023-03-24
add conversions between Timespec and Duration

FrankReh:
Like the nix crate does:

On Linux, the std::time::Duration and __kernel_timespec should be trivially interchangeable.

This implements the std::convert::From trait in each direction.


quininer (approved):
I don't want to add too many auxiliary methods to these types so that they keep simple and low-level. but if you want, i will approve to this, because it looks useful.

FrankReh:
I'm assuming you would know that the compiler will optimize the TimeSpec sequence of calls
```rust
types::Timespec::new().sec(dur.as_secs()).nsec(dur.subsec_nanos())
```
because I myself am not sure. So I don't think my `From` impl's actually add any value. I realize there is no reason to convert back from a TimeSpec to a Duration. So I'm going to close this. Thanks for considering it though.

PR #223, opened 2023-03-24
adds newer timeout bitflags

FrankReh:


src/opcode.rs
```diff
@@ -549,7 +549,15 @@ opcode!(
         /// `count` may contain a completion event count.
         count: u32 = 0,
 
-        /// `flags` may contain [types::TimeoutFlags::ABS] for an absolute timeout value, or 0 for a relative timeout.
+        /// The default behavior is to treat the timespec as a relative time interval. `flags` may
+        /// contain [`types::TimeoutFlags::ABS`] to indicate the timespec represents an absolute
```
quininer, line 553: should we put these docs to `TimeoutFlags`?

quininer (changes requested):

src/opcode.rs
FrankReh: I would have said no, but I think this is your way of kindly asking me to move them. So I'll move them. But it is a little awkward in the new location because the flags are relevant to Timeout and TimeoutUpdate but not to TimeoutRemove.

src/types.rs
```diff
@@ -77,10 +77,28 @@ pub struct Fixed(pub u32);
...
+    /// contain [`types::TimeoutFlags::ABS`] to indicate the timespec represents an absolute
+    /// time. When an absolute time is being specified, the kernel will use its monotonic clock
+    /// unless one of the following flags is set (they may not both be set):
+    /// [`types::TimeoutFlags::BOOTTIME`] or [`types::TimeoutFlags::REALTIME`].
+    ///
+    /// The default behavior when the timeout expires is to return a CQE with -libc::ETIME in
+    /// the res field. To change this behavior to have zero returned, include
+    /// [`types::TimeoutFlags::ETIME_SUCCESS`].
     pub struct TimeoutFlags: u32 {
         const ABS = sys::IORING_TIMEOUT_ABS;
```
quininer, line 91: I'm considering whether it would be better to add docs for each field.

src/opcode.rs
quininer: We can explain its effect scope in docs. or as another MR said, we can remove it from TimeoutRemove.

src/types.rs
FrankReh: I am not the right person to do justice to the individual field comments. There are many cases in man pages and even Rust documentation where too much documentation is given per field and the holistic documentation about the whole set of fields is missing or is too short. -- Just the way my brain works. Thinking about it now, I guess I was happy to try to provide a holistic description when it was part of the Timeout `flags` parameter.

So let's leave this open then and wait for someone with better or different explanation skills to make the per field documentation make sense. If no-one steps up to make the comment changes, I'm okay with closing this PR too.

quininer: It's okay, it's good enough

quininer (approved):

PR #224, opened 2023-03-24
opcode: add TimeoutUpdate

FrankReh:


src/opcode.rs
```diff
@@ -576,6 +576,9 @@ opcode!(
     pub struct TimeoutRemove {
         user_data: { u64 },
         ;;
+        /// N.B. While the compiler will allow it, it is a mistake to include the
```
quininer, line 579: There seems no need to emphasize that this is a mistake, and it seems reasonable for use TIMEOUT_UPDATE flags to have different behavior.

quininer (changes requested):

src/opcode.rs
FrankReh: I will add a commit that removes this comment. But I think you or I are still under a misconception.

I think you should let me remove the TIMOUT_UPDATE bit flag entirely from the definition. There is no reason a user of the library should be passing it. It seems clear to me from how it is used in the liburing library that it is only intended to differentiate the remove function from the update function when using the IORING_OP_TIMEOUT_REMOVE.

In the liburing case, they have a different API function so their users never use that flag either. And in our case, we have a different op, TimeoutUpdate, so our users would never use that flag either.

Would you accept my removing the flag with this PR?

FrankReh: > and it seems reasonable for use TIMEOUT_UPDATE flags to have different behavior.

Just to try and clarify my reasoning even further, maybe I will comment on this. The user can't get different behavior by setting this flag, because they would need to pass a new Timespec at the same time, and the TimeoutRemove doesn't accept a new Timespec. So we shouldn't be confusing the user with a flag they can't use.

quininer: I agree that we can remove it, and if flag have another feature in future, we can always add it back.

quininer (approved):
LGTM

PR #225, opened 2023-03-25
Add support for IORING_REGISTER_SYNC_CANCEL

gardnervickers:
Adds a new Submitter method, register_sync_cancel which can be used to synchronously cancel operations.
Additionally, this adds AsyncCancelFlags used to configure cancellation.

I didn't integrate the `AsyncCancelFlags` with the `opcode::AsyncCancel` in this PR, mostly because I haven't needed it but I can either do that in this PR or a follow-up PR.

### DIFF
--- a/io-uring-test/src/tests/register_sync_cancel.rs
+++ b/io-uring-test/src/tests/register_sync_cancel.rs
@@ -32,7 +32,6 @@ pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>
     // 3 operations, canceled by ANY. Reads from fd_2.
     const USER_DATA_3: u64 = 45u64;
 
-    // Start reads for user_data_0
     let mut buf = [0u8; 32];
     for i in 0..8 {
         let entry;
@@ -55,11 +54,10 @@ pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>
         }
         unsafe { ring.submission().push(&entry.into()).unwrap() };
     }
-    // Submit all 7 operations.
+    // Submit all 8 operations.
     assert_eq!(8, ring.submit()?);
 
     // Cancel the first operation by user_data.
-
     ring.submitter()
         .register_sync_cancel(None, CancelBuilder::new().user_data(USER_DATA_0))?;
     let completions = wait_get_completions(ring, 1).unwrap();
@@ -89,7 +87,7 @@ pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>
         assert_eq!(completion.result(), -libc::ECANCELED);
     }
 
-    // Cancel all of the remaining requests.
+    // Cancel all of the remaining requests, should be 3 outstanding.
     ring.submitter()
         .register_sync_cancel(None, CancelBuilder::new())?;
     let completions = wait_get_completions(ring, 1).unwrap();
@@ -110,6 +108,7 @@ pub fn test_register_sync_cancel_any<S: squeue::EntryMarker, C: cqueue::EntryMar
         test;
         test.probe.is_supported(opcode::SendZc::CODE);
     );
+
     // Test that CancelBuilder::new().all() cancels all requests in the ring.
     let fd_1 = get_eventfd();
     const START_USER_DATA: u64 = 47u64;
@@ -124,7 +123,11 @@ pub fn test_register_sync_cancel_any<S: squeue::EntryMarker, C: cqueue::EntryMar
     // Submit all 3 operations.
     assert_eq!(3, ring.submit()?);
 
-    // Cancel all of the requests.
+    // Cancel all of the requests by supplying CancelBuilder::new().all(). This should result in
+    // flags IORING_ASYNC_CANCEL_ALL | IORING_ASYNC_CANCEL_ANY, which match all in-flight requsts.
+    //
+    // Note: IORING_ASYNC_CANCEL_ALL makes no difference here, IORING_ASYNC_CANCEL_ANY will behave
+    //       the same without it. This test just verifies that the builder works as expected.
     ring.submitter()
         .register_sync_cancel(None, CancelBuilder::new().all())?;
 
@@ -157,7 +160,7 @@ pub fn test_register_sync_cancel_unsubmitted<S: squeue::EntryMarker, C: cqueue::
 
     let fd_1 = get_eventfd();
 
-    // Test that we can cancel operations which have not yet been submitted.
+    // Test that we see an error when attempting to cancel operations which have not yet been submitted.
     const USER_DATA: u64 = 47u64;
 
     let mut buf = [0u8; 32];
@@ -166,7 +169,7 @@ pub fn test_register_sync_cancel_unsubmitted<S: squeue::EntryMarker, C: cqueue::
         .user_data(USER_DATA);
     unsafe { ring.submission().push(&entry.into()).unwrap() };
 
-    // Cancel the operation by user_data, we haven't submitted anything yet.
+    // Cancel the operation by user_data, we haven't submitted anything yet. We should get an error.
     let result = ring
         .submitter()
         .register_sync_cancel(None, CancelBuilder::new().user_data(USER_DATA));
--- a/io-uring-test/src/tests/register_sync_cancel.rs
+++ b/io-uring-test/src/tests/register_sync_cancel.rs
@@ -1,10 +1,9 @@
 use std::io;
-use std::time::{Duration, Instant};
 
 use io_uring::cqueue;
 use io_uring::opcode;
-use io_uring::squeue::{self, Flags};
-use io_uring::types::{self, AsyncCancelFlags, Timespec};
+use io_uring::squeue;
+use io_uring::types::{self, AsyncCancelFlags};
 use io_uring::IoUring;
 
 use crate::Test;
@@ -90,7 +89,7 @@ pub fn test_register_sync_cancel_unsubmitted<S: squeue::EntryMarker, C: cqueue::
     const USER_DATA: u64 = 42u64;
 
     let mut buf = [0u8; 32];
-    let mut entry = opcode::Read::new(types::Fd(0), buf.as_mut_ptr(), 32).build();
+    let entry = opcode::Read::new(types::Fd(0), buf.as_mut_ptr(), 32).build();
     unsafe { ring.submission().push(&entry.into()).unwrap() };
 
     // Cancel the operation by user_data, we haven't submitted anything yet.
--- a/io-uring-test/src/tests/register_sync_cancel.rs
+++ b/io-uring-test/src/tests/register_sync_cancel.rs
@@ -1,0 +1,106 @@
+use std::io;
+use std::time::{Duration, Instant};
+
+use io_uring::cqueue;
+use io_uring::opcode;
+use io_uring::squeue::{self, Flags};
+use io_uring::types::{self, AsyncCancelFlags, Timespec};
+use io_uring::IoUring;
+
+use crate::Test;
+
+pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> io::Result<()> {
+    require!(
+        test; // We need at least 5.19 for these tests, UringCmd16 is a proxy for that.
+        test.probe.is_supported(opcode::UringCmd16::CODE);
+    );
+    // Single op, canceled by the user_data
+    const USER_DATA_0: u64 = 42u64;
+    // 2 operations, canceled by the user_data
+    const USER_DATA_1: u64 = 43u64;
+    // 2 operations, canceled by the fd.
+    const USER_DATA_2: u64 = 44u64;
+
+    // Start reads for user_data_0
+    let mut buf = [0u8; 32];
+    for i in 0..5 {
+        let mut entry = opcode::Read::new(types::Fd(0), buf.as_mut_ptr(), 32).build();
+
+        if i == 0 {
+            entry = entry.user_data(USER_DATA_0);
+        } else if (1..3).contains(&i) {
+            entry = entry.user_data(USER_DATA_1);
+        } else if (3..5).contains(&i) {
+            entry = entry.user_data(USER_DATA_2);
+        }
+        unsafe { ring.submission().push(&entry.into()).unwrap() };
+    }
+    // Submit all 5 operations.
+    assert_eq!(5, ring.submit()?);
+
+    // Cancel the first operation by user_data
+    ring.submitter()
+        .register_sync_cancel(USER_DATA_0, -1, None, AsyncCancelFlags::empty())?;
+    ring.submitter().submit_and_wait(1)?;
+    let cqe: cqueue::Entry = ring.completion().next().unwrap().into();
+    assert_eq!(
+        cqe.result(),
+        -libc::ECANCELED,
+        "operation should have been canceled"
+    );
+    assert_eq!(cqe.user_data(), USER_DATA_0);
+
+    // Cancel the second two operations by their user data flags.
+    ring.submitter()
+        .register_sync_cancel(USER_DATA_1, -1, None, AsyncCancelFlags::ALL)?;
+    ring.submitter().submit_and_wait(2)?;
+    let cqe1: cqueue::Entry = ring.completion().next().unwrap().into();
+    let cqe2: cqueue::Entry = ring.completion().next().unwrap().into();
+    assert_eq!(cqe1.result(), cqe2.result());
+    assert_eq!(cqe1.result(), -libc::ECANCELED);
+
+    // Cancel the last two operations by their fd.
+    ring.submitter().register_sync_cancel(
+        USER_DATA_2,
+        0,
+        None,
+        AsyncCancelFlags::FD | AsyncCancelFlags::ALL,
+    )?;
+    ring.submitter().submit_and_wait(2)?;
+    let cqe1: cqueue::Entry = ring.completion().next().unwrap().into();
+    let cqe2: cqueue::Entry = ring.completion().next().unwrap().into();
+    assert_eq!(cqe1.result(), cqe2.result());
+    assert_eq!(cqe1.result(), -libc::ECANCELED);
+
+    Ok(())
+}
+
+pub fn test_register_sync_cancel_unsubmitted<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> io::Result<()> {
+    require!(
+        test; // We need at least 5.19 for these tests, UringCmd16 is a proxy for that.
+        test.probe.is_supported(opcode::UringCmd16::CODE);
+    );
+    // Test that we can cancel operations which have not yet been submitted.
+    const USER_DATA: u64 = 42u64;
+
+    let mut buf = [0u8; 32];
+    let mut entry = opcode::Read::new(types::Fd(0), buf.as_mut_ptr(), 32).build();
+    unsafe { ring.submission().push(&entry.into()).unwrap() };
+
+    // Cancel the operation by user_data, we haven't submitted anything yet.
+    ring.submitter()
+        .register_sync_cancel(USER_DATA, -1, None, AsyncCancelFlags::empty())?;
+
+    ring.submitter().submit_and_wait(1)?;
+    let cqe1: cqueue::Entry = ring.completion().next().unwrap().into();
+    assert_eq!(cqe1.result(), -libc::ECANCELED);
+    assert_eq!(cqe1.user_data(), USER_DATA);
+
+    Ok(())
+}
--- a/io-uring-test/src/tests/register_sync_cancel.rs
+++ b/io-uring-test/src/tests/register_sync_cancel.rs
@@ -83,7 +83,7 @@ pub fn test_register_sync_cancel_unsubmitted<S: squeue::EntryMarker, C: cqueue::
         test.probe.is_supported(opcode::SendZc::CODE);
     );
     // Test that we can cancel operations which have not yet been submitted.
-    const USER_DATA: u64 = 42u64;
+    const USER_DATA: u64 = 45u64;
 
     let mut buf = [0u8; 32];
     let entry = opcode::Read::new(types::Fd(0), buf.as_mut_ptr(), 32)
--- a/io-uring-test/src/tests/register_sync_cancel.rs
+++ b/io-uring-test/src/tests/register_sync_cancel.rs
@@ -1,4 +1,5 @@
 use std::io;
+use std::mem;
 use std::os::fd::AsRawFd;
 use std::os::fd::FromRawFd;
 use std::os::fd::OwnedFd;
@@ -22,6 +23,11 @@ pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>
     );
     let fd_1 = get_eventfd();
     let fd_2 = get_eventfd();
+    let fixed_fd_3 = get_eventfd();
+
+    let mut registered_files = [0; 1];
+    registered_files[0] = fixed_fd_3.as_raw_fd();
+    ring.submitter().register_files(&registered_files)?;
 
     // Single op, canceled by the user_data. Reads from fd_1
     const USER_DATA_0: u64 = 42u64;
@@ -29,11 +35,14 @@ pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>
     const USER_DATA_1: u64 = 43u64;
     // 2 operations, canceled by the fd. Reads from fd_1
     const USER_DATA_2: u64 = 44u64;
-    // 3 operations, canceled by ANY. Reads from fd_2.
+    // 3 operations, reading from fixed_fd_3.
     const USER_DATA_3: u64 = 45u64;
+    // 3 operations, canceled by ANY. Reads from fd_2.
+    const USER_DATA_4: u64 = 46u64;
 
     let mut buf = [0u8; 32];
-    for i in 0..8 {
+    let mut nr_submitted = 0;
+    for i in 0..11 {
         let entry;
         if i == 0 {
             entry = opcode::Read::new(types::Fd(fd_1.as_raw_fd()), buf.as_mut_ptr(), 32)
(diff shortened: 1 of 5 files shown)


src/submit.rs
```diff
@@ -476,4 +477,44 @@ impl<'a> Submitter<'a> {
...
+    ///
+    /// [`AsyncCancelFlags`](types::AsyncCancelFlags) may be supplied to indicate the strategy for
+    /// matching operations to cancel.
+    ///
+    /// Available since 6.0.
+    pub fn register_sync_cancel(
+        &self,
+        user_data: u64,
+        fd: RawFd,
+        timeout: Option<Duration>,
```
gardnervickers, line 492: I had a few questions with the public API here:

1. Do we want to take a `Duration` or a `Timespec`? 
2. Users can cancel based on either a "regular" FD or a fixed FD via the `AsyncCancelFlags`. Should we remove the distinction in the flags and automatically pick the correct flag based on the `fd` argument? I think that would mean that we'd take `impl sealed::UseFixed` instead of a raw fd.

io-uring-test/src/tests/register_sync_cancel.rs
```diff
@@ -0,0 +1,105 @@
...
+use io_uring::IoUring;
+
+use crate::Test;
+
+pub fn test_register_sync_cancel<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> io::Result<()> {
+    require!(
+        test; // We need at least 5.19 for these tests, UringCmd16 is a proxy for that.
```
gardnervickers, line 16: Hmm this is wrong I think we need to probe for 6.0

gardnervickers: I switched this to probe for `opcode::SendZc::CODE` since that seems to be available only on 6.0.

src/submit.rs
quininer: 1. I think using `Timespec` has more consistency.
2. for consistency reasons, `UseFixed` would be good.
3. I think `None` for infinite is good because `Timespec::new()` can easily express zero.

io-uring-test/src/tests/register_sync_cancel.rs
```diff
@@ -0,0 +1,105 @@
...
+        .register_sync_cancel(USER_DATA_1, -1, None, AsyncCancelFlags::ALL)?;
+    ring.submitter().submit_and_wait(2)?;
+    let cqe1: cqueue::Entry = ring.completion().next().unwrap().into();
+    let cqe2: cqueue::Entry = ring.completion().next().unwrap().into();
+    assert_eq!(cqe1.result(), cqe2.result());
+    assert_eq!(cqe1.result(), -libc::ECANCELED);
+
+    // Cancel the last two operations by their fd.
+    ring.submitter().register_sync_cancel(
+        USER_DATA_2,
```
FrankReh, line 66: When cancelling by FD, passing the user_data should not be necessary, so  probably should not be part of the test example.

src/submit.rs
```diff
@@ -476,4 +477,44 @@ impl<'a> Submitter<'a> {
...
         .map(drop)
     }
+
+    /// Performs a synchronous cancellation request, similiar to [AsyncCancel](crate::opcode::AsyncCancel),
+    /// except that it completes synchronously.
+    ///
+    /// [`AsyncCancelFlags`](types::AsyncCancelFlags) may be supplied to indicate the strategy for
+    /// matching operations to cancel.
+    ///
+    /// Available since 6.0.
```
FrankReh, line 487: This comment block should probably indicate what values to use for the optional values. Like 0 for user_data and -1 for fd. The -1 for fd wasn't obvious to me until I saw the test case use it. Passing zero when a library like tokio-uring uses slab indexes where zero is very common would make me nervous but a little testing with the correct flag set would resolve that it was okay for me. I might just pass the max value allowed anyway making it more obvious.

It's a little strange to me that only the timeout parameter actually would take an Option<T> but that doesn't bother quininer so I am in the minority, so just a comment, not a request for change. Other possibilities were all three are options, or the first two are a single enum with two invariants. But I understand this library's job is to provide bindings, not necessarily the strictest API surface for the promised semantics.

src/submit.rs
```diff
@@ -476,4 +477,44 @@ impl<'a> Submitter<'a> {
...
+    pub fn register_sync_cancel(
+        &self,
+        user_data: u64,
+        fd: RawFd,
+        timeout: Option<Duration>,
+        flags: types::AsyncCancelFlags,
+    ) -> io::Result<()> {
+        // Default to infinite timeout if a timeout is not provided.
+        let mut timespec = sys::__kernel_timespec {
```
FrankReh, line 496: Interesting that because this is a synchronous call, there is no issue of having to hold the Timespec data in a 'static location as in the Timeout opcode, or as there would be in the AsyncCancel opcode (if that opcode handled more of the underlying feature set).

As a result, I would have opted for std::time::Duration being the type here but quininer already weighed in that consistency of type, even though one is by pointer and the other by value, was to be preferred.

FrankReh (commented):
This is very nice. I couldn't help but express some opinions but my overall opinion is this is very well done.

io-uring-test/src/tests/register_sync_cancel.rs
```diff
@@ -0,0 +1,105 @@
...
+    );
+    // Test that we can cancel operations which have not yet been submitted.
+    const USER_DATA: u64 = 42u64;
+
+    let mut buf = [0u8; 32];
+    let entry = opcode::Read::new(types::Fd(0), buf.as_mut_ptr(), 32).build();
+    unsafe { ring.submission().push(&entry.into()).unwrap() };
+
+    // Cancel the operation by user_data, we haven't submitted anything yet.
+    ring.submitter()
```
gardnervickers, line 96: This should be failing because we have't submitted the operation, and sync cancel will only work for operations which have been submitted.

FrankReh: That's odd. Is it being run in parallel with the other test on the same ring? Have you tried with a different USER_DATA value?

src/submit.rs
gardnervickers: Thanks! I agree the arguments are pretty confusing at the moment so any way we can make them safer or more obvious sounds good to me. 

> The -1 for fd wasn't obvious to me
I believe you can pass any value in for the FD as long as you do not set the `AsyncCancelFlag::FD/FD_FIXED` flag. I was just passing in `-1` to try and make it obvious that we were not canceling by the FD we were reading from in the tests. 

Maybe we could encode the match criteria in a builder-like thing. 

For example:
```
types::MatchOn::fd(types::Fd(4)).and_user_data(42).match_any();
```
This would match any requests where either the FD = 42 or the user_data = 42.

With this, the signature would be
```
pub fn register_sync_cancel(&self, match_on: types::MatchOn, timeout: Option<Timespec>) -> io::Result<()>;
```
This might be the way to go because we can't really support `Option<impl UseFixed>` directly in the API, because the compiler will complain about the concrete type being unknown if we supply `None` for the FD. 

I'll try this out and we can see if it's worth adding another builder type. c

io-uring-test/src/tests/register_sync_cancel.rs
gardnervickers: Sorry, I meant that this test should be checking for failure rather than success. It was failing previously, but it seems that CI might not be running these tests so it wasn't picked up until I ran it locally.

gardnervickers:
Thanks for the review @quininer and @FrankReh. I pushed a new set of changes which include a builder for constructing match criteria for the `register_sync_cancel` call.

The main motivation for it was that taking `Option<impl UseFixed>` in `register_sync_cancel` would require users to specify a concrete type even if they supply `None` for an FD. With the `MatchOn` builder, we can take `impl UseFixed` and just extract the `sealed::Target` from it.

Additionally, I think the `MatchOn` builder makes it a bit easier to supply the correct set of arguments, since users don't set the cancellation flags directly now. Please let me know what you think, I wasn't sure if it was adding too much extra complexity for this crate or not.

io-uring-test/src/tests/register_sync_cancel.rs
gardnervickers: Thanks, this should be addressed with the new `MatchOn` builder now.

io-uring-test/src/tests/register_sync_cancel.rs
FrankReh: Well that makes sense. I was wondering how that was going to be worked out. I didn't even think that I might have been misunderstanding what was meant.

src/submit.rs
```diff
@@ -476,4 +477,70 @@ impl<'a> Submitter<'a> {
         )
         .map(drop)
     }
+
+    /// Performs a synchronous cancellation request, similiar to [AsyncCancel](crate::opcode::AsyncCancel),
+    /// except that it completes synchronously.
+    ///
+    /// Cancelation can target a specific request, or all requests matching a specific criteria. The
```
FrankReh, line 484: small nit. `Cancellation` has two els. Appears misspelled several times here.

src/submit.rs
```diff
@@ -476,4 +477,70 @@ impl<'a> Submitter<'a> {
...
+    /// [MatchOn](types::MatchOn) builder supports describing the match criteria for cancelation.
+    ///
+    /// An optional `timeout` can be provided to specify how long to wait for matched requests to be
+    /// canceled. If no timeout is provided, the default is to wait indefinitely.
+    ///
+    /// ### Errors
+    ///
+    /// An error is returned if no requests match the provided match criteria. If a timeout is supplied,
+    /// and the timeout expires before all requests have been canceled, then an error will be returned.
```
FrankReh, line 493: Given that a timeout return would be important to recognize, I think it useful to call out that it returns ETIME (or whatever the io::Error module converts that too).

src/types.rs
```diff
@@ -475,3 +506,209 @@ impl<'buf> RecvMsgOut<'buf> {
         self.header.flags
     }
 }
+
+/// [MatchOn] is a builder for constructing match criteria for request cancelation.
```
FrankReh, line 510: Cancellation. I see that `cancelation` can be said to be technically correct, but it appears very infrequently in software. Now that I've seen it used here a few times, I've grown to like it, but still,
it's not used in the `tokio` library while the form with two els is used often.

src/types.rs
```diff
@@ -475,3 +506,209 @@ impl<'buf> RecvMsgOut<'buf> {
...
+    /// use io_uring::types;
+    /// // Match any request where the user_data or fd is 42.
+    /// types::MatchOn::user_data(42).and_fd(types::Fixed(42)).match_any();
+    /// ```
+    pub fn match_any(mut self) -> Self {
+        self.multiple = MatchMultiple::Any;
+        self
+    }
+
+    /// Additionally, match a request with the given user_data field.
```
FrankReh, line 617: Copy and paste error in this comment?

This is about `and_fd`.

src/types.rs
```diff
@@ -475,3 +506,209 @@ impl<'buf> RecvMsgOut<'buf> {
...
+    /// ### Example
+    ///
+    /// ```
+    /// use io_uring::types;
+    /// // Match a single request where the user_data is
+    /// // 42 or the fd is 4.
+    /// types::MatchOn::user_data(42).and_fd(types::Fixed(4));
+    /// ```
+    ///
+    /// Note: This will overwrite any previous user_data match criteria.
```
FrankReh, line 628: not `user_data`

src/types.rs
```diff
@@ -475,3 +506,209 @@ impl<'buf> RecvMsgOut<'buf> {
...
+    pub fn and_fd(mut self, fd: impl sealed::UseFixed) -> Self {
+        if matches!(self.multiple, MatchMultiple::All) {
```
FrankReh, line 630: I don't follow the reasoning behind this test for comparison to All. One would still want the ability to match against all for a given fd.

src/types.rs
```diff
@@ -475,3 +506,209 @@ impl<'buf> RecvMsgOut<'buf> {
...
+    /// ### Example
+    ///
+    /// ```
+    /// use io_uring::types;
+    /// // Match a single request where the user_data is
+    /// // fd 4 or the user_data is 42.
+    /// types::MatchOn::fd(types::Fd(4)).and_user_data(42);
+    /// ```
+    pub fn and_user_data(mut self, user_data: u64) -> Self {
+        if matches!(self.multiple, MatchMultiple::All) {
```
FrankReh, line 648: Same.

src/types.rs
```diff
@@ -475,3 +506,209 @@ impl<'buf> RecvMsgOut<'buf> {
...
+    pub(crate) user_data: Option<u64>,
+    pub(crate) fd: Option<sealed::Target>,
+    multiple: MatchMultiple,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum MatchMultiple {
+    /// Match all requests. This indiscriminately cancels all requests,
+    /// regardless of user_data or fd.
+    All,
```
FrankReh, line 531: You and I might have different understandings of the `all` match.

From the man page:
```
IORING_ASYNC_CANCEL_ALL
              Cancel all requests that match the given criteria, rather than just  can‐
              celing the first one found. Available since 5.19.
```

In fact, we might have different interpretations for what combinations are allowed. The liburing helper functions allow matching on user_data or fd, not both.

The `all` means all with the same user_data or all with the same fd.

Or the `any` flag means don't bother looking at the user_data, just match any, (and I'm assuming) if `any` and `all` are specified, it means all in-flight operations. It would be very odd to ask for `any` and just have the first matched, but that would seem to be allowed.

src/submit.rs
```diff
@@ -476,4 +477,70 @@ impl<'a> Submitter<'a> {
...
+            }
+        };
+
+        let flags = match_on.flags();
+        let user_data = match_on.user_data.unwrap_or(0);
+        let fd = match_on
+            .fd
+            .map(|fd| match fd {
+                types::sealed::Target::Fd(fd) => fd,
+                types::sealed::Target::Fixed(fd) => fd as i32,
```
FrankReh, line 526: Maybe the register version of this can't cancel a fixed file descriptor? I will submit a question to the liburing repo.

FrankReh (changes requested):

src/types.rs
FrankReh:
```
CancelBuilder::new() // match on any
user_data(mut self, user_data) -> Self // match on user_data rather than any
fd(mut self, fd) -> Self // match on fd rather than any
all(mut self) -> Self // match on all rather than just the first.
```
So I think the builder impl could have just the four functions.

gardnervickers: Thanks, I think I was swapping around ALL and ANY here.

The following statements seem true based on my testing and read of the manpage.
1. Supplying a FD argument requires setting either **one of** sys::IORING_ASYNC_CANCEL_FD or sys::IORING_ASYNC_CANCEL_FD_FIXED.
2. Supplying a FD argument "overrides" the userdata argument. Testing shows that if you supply both, the FD argument takes precedence. 
3. `IORING_ASYNC_CANCEL_ALL` will cancel all requests which match either the FD argument or the user data argument.
4. `IORING_ASYNC_CANCEL_ANY` will cancel all operations, regardless of the other flags that were set. 

This would make the API you suggested output flags like the following:

- `CancelBuilder::new()` -> IORING_ASYNC_CANCEL_ANY: Cancel everything
- `CancelBuilder::user_data()` -> no flags: Cancel one request matching user_data.
- `CancelBuilder::fd()` -> IORING_ASYNC_CANCEL_FD: Cancel one request matching the FD or the fixed FD for FD_FIXED.
- `CancelBuilder::all()` previous flags | IORING_ASYNC_CANCEL_ALL: Cancel all matching requests. 


I'll implement this and try to beef up the test suite to check these combinations. Thanks again!

FrankReh: yes, I concur with your descriptions. Bear in mind quininer hasn't sanctioned any of this as their timezone is completely different. But the Timespec has a small builder associated with it, and the semantics of the cancel are interesting enough that I would say the builder is worth it here too.

As for your point 1, I wouldn't be surprised if IORING_ASYNC_CANCEL_FD was supposed to be set for either type, regular and fixed, and only the *FD_FIXED flag should be added to make it the fixed version. But maybe the kernel accepts it both ways.

io-uring-test/src/tests/register_sync_cancel.rs
```diff
@@ -0,0 +1,213 @@
...
+    test: &Test,
+) -> io::Result<()> {
+    require!(
+        test;
+        test.probe.is_supported(opcode::SendZc::CODE);
+    );
+
+    let fd_1 = get_eventfd();
+
+    // Test that we can cancel operations which have not yet been submitted.
```
FrankReh: Should this read something like: Try to cancel an operation which has not yet been submitted?

src/types.rs
```diff
@@ -91,6 +91,37 @@ bitflags! {
     }
 }
 
+bitflags! {
+    /// Options for [`AsyncCancel`](super::AsyncCancel) and
+    /// [`Submitter::register_sync_cancel`](super::Submitter::register_sync_cancel).
+    pub struct AsyncCancelFlags: u32 {
```
FrankReh, line 97: Any reason to make this public now?

src/types.rs
```diff
@@ -475,3 +506,150 @@ impl<'buf> RecvMsgOut<'buf> {
         self.header.flags
     }
 }
+
+/// [CancelBuilder] constructs match criteria for request cancellation.
+///
+/// The default behavior of the [CancelBuilder] is cancel all in-flight requests,
+/// however this can be refined to cancel only requests which reference a file descriptor
+/// or have a specific `user_data` value.
```
FrankReh, line 514: This is a very good start to the comment block.

src/types.rs
```diff
@@ -475,3 +506,150 @@ impl<'buf> RecvMsgOut<'buf> {
...
+        // Unset the ANY flag because we want to refine the match criteria to a subset of requests.
+        self.flags.set(AsyncCancelFlags::ANY, false);
+        let target = fd.into();
+        self.user_data = None;
+        match target {
+            sealed::Target::Fd(_) => {
+                self.flags.set(AsyncCancelFlags::FD, true);
+            }
+            sealed::Target::Fixed(_) => {
+                self.flags.set(AsyncCancelFlags::FD_FIXED, true);
```
FrankReh, line 590: For this, I would have thought both need to be set, but if you've tested and it works this way, then that's fine.

FrankReh (changes requested):
Very nice. I think the use of the builder API in the test cases shows its value. It makes the API much easier to use than giving the user a bunch of flags and letting them figure it out through trial and error.

I think my only request was for a minor comment change about 'all'. And maybe an earlier comment about the error that is returned when the timeout expires.

But those are such small nits at this point. I would approve this. It's nice.

src/types.rs
```diff
@@ -475,3 +506,152 @@ impl<'buf> RecvMsgOut<'buf> {
...
+///
+/// Additionally, [CancelBuilder::all] can be set to match multiple requests.
+///
+/// ### Examples
+///
+/// ```
+/// use io_uring::types::{CancelBuilder, Fd, Fixed};
+///
+/// // Match all in-flight requests.
+/// CancelBuilder::new();
```
quininer, line 524: This seems counter-intuitive, maybe it should be called `CancelBuilder::any()` ?

src/types.rs
```diff
@@ -475,3 +506,152 @@ impl<'buf> RecvMsgOut<'buf> {
...
+    pub(crate) user_data: Option<u64>,
+    pub(crate) fd: Option<sealed::Target>,
+}
+
+impl CancelBuilder {
+    /// Create a new [CancelBuilder].
+    ///
+    /// The default behavior of the [CancelBuilder] is cancel all in-flight requests.
+    /// This behavior can be further refined by calling the builder methods on the [CancelBuilder].
+    pub fn new() -> Self {
```
quininer, line 550: trying to make it const would be nice

src/types.rs
gardnervickers: Thanks for pointing this out! I checked again, I think you are right that we need both CANCEL_FD and CANCEL_FD_FIXED.

The checks for CANCEL_FD_FIXED seem to be gated on CANCEL_FD.
https://github.com/torvalds/linux/blob/3a93e40326c8f470e71d20b4c42d36767450f38f/io_uring/cancel.c#L177-L179

I updated the PR to include a test for fixed fd cancellation, and it fails unless we include the CANCEL_FD_FIXED flag.

FrankReh:
There may still be some issues outstanding but I wanted to say the model you chose for the unit test and the use of stdin to have operations blocked on is very nice. I don't remember seeing that in the unit tests before. I suspect it will be copied for other unit tests down the line.

src/types.rs
```diff
@@ -475,3 +506,157 @@ impl<'buf> RecvMsgOut<'buf> {
...
+        }
+    }
+
+    /// Refine the match criteria to match only requests which contain the
+    /// provided `user_data`.
+    ///
+    /// This is mutually exclusive with [CancelBuilder::fd](#method.fd).
+    pub fn user_data(mut self, user_data: u64) -> Self {
+        // Unset the ANY flag because we want to refine the match criteria to a subset of requests.
+        self.flags.set(AsyncCancelFlags::ANY, false);
```
quininer, line 565: I'm a bit curious which is better to use insert/remove vs set .

src/types.rs
gardnervickers: `CancelBuilder::any()` sounds good to me. Should we likewise make the `.all()` method a constructor too? I don't think users will ever want to call `CancelBuilder::any().all()`.

src/submit.rs
```diff
@@ -476,4 +477,74 @@ impl<'a> Submitter<'a> {
...
+    /// Available since 6.0.
+    pub fn register_sync_cancel(
+        &self,
+        timeout: Option<Timespec>,
+        builder: CancelBuilder,
+    ) -> io::Result<()> {
+        let timespec = match timeout {
+            Some(ref ts) => {
+                // Safety: Timespec is repr(transparent) over __kernel_timespec, so the cast is safe.
+                unsafe { *(ts as *const Timespec as *const sys::__kernel_timespec) }
```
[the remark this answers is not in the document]
gardnervickers: Should I access the inner `sys::__kernel_timespec`  from `Timespec` directly or is there something else you had in mind?

src/types.rs
quininer: I think it would be nice to have `all()`.

src/submit.rs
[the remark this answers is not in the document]
quininer: You can directly access the inner.

src/types.rs
gardnervickers: I'm not sure, tbh I didn't realize insert/remove existed but it seems a bit clearer to me, so I switched to using that instead of set.

src/types.rs
gardnervickers: @quininer @FrankReh I changed the builder yet again, I think it might be a bit easier to use now. We allow users to do the following:

```
| op                                 | action                            |
|------------------------------------+-----------------------------------|
| CancelBuilder::any()               | Cancel all in-flight ops          |
| CancelBuilder::any().all()         | Cancel all in-flight ops          |
| CancelBuilder::user_data(42)       | Cancel the first op with udata=42 |
| CancelBuilder::user_data(42).all() | Cancel all ops with udata=42      |
| CancelBuilder::fd(1)               | Cancel the first op with fd=1     |
| CancelBuilder::fd(1).all()         | Cancel all ops with fd=1          |
```

gardnervickers:
> There may still be some issues outstanding but I wanted to say the model you chose for the unit test and the use of stdin to have operations blocked on is very nice. I don't remember seeing that in the unit tests before. I suspect it will be copied for other unit tests down the line.


Thanks, I actually ended up switching away from using stdin in favor of creating an eventfd, partly because I was worried about using a shared resource in the tests, and partly because I wasn't sure if there would be a problem using stdin with fixed files.

src/types.rs
```diff
@@ -4,7 +4,7 @@ pub(crate) mod sealed {
     use super::{Fd, Fixed};
     use std::os::unix::io::RawFd;
 
-    #[derive(Debug)]
+    #[derive(Debug, PartialEq, Eq)]
```
quininer, line 7: If there is an alternative, I would like to avoid it.

src/types.rs
quininer: I don't have a strong opinion, I think it's all good. :)

src/types.rs
```diff
@@ -475,3 +506,126 @@ impl<'buf> RecvMsgOut<'buf> {
...
+///
+/// The [CancelBuilder] can be used to selectively cancel one or more requests
+/// by user_data, fd, fixed fd, or unconditionally.
+///
+/// ### Examples
+///
+/// ```
+/// use io_uring::types::{CancelBuilder, Fd, Fixed};
+///
+/// // Match all in-flight requests.
```
FrankReh, line 520: Probably should read 'Match any ...'. It will just cancel the first one found. The `All` bit isn't set. Maybe given an any().all() doc example too.

src/types.rs
```diff
@@ -475,3 +506,126 @@ impl<'buf> RecvMsgOut<'buf> {
...
+/// CancelBuilder::any();
+///
+/// // Match a single request with user_data = 42.
+/// CancelBuilder::user_data(42);
+///
+/// // Match a single request with fd = 42.
+/// CancelBuilder::fd(Fd(42));
+///
+/// // Match a single request with fixed fd = 42.
+/// CancelBuilder::fd(Fixed(42));
```
FrankReh, line 530: This could be a fixed_fd method.

FrankReh (changes requested):
Yes, I think the change to the builder make sense. Now there are a few strict combinations allowed.

{any|user_data|fd} (.all)

You could split fd into two: fd and fixed_fd, or some other set of names.

src/types.rs
gardnervickers: https://github.com/tokio-rs/io-uring/pull/225/files#diff-3cb81e12d8f6c5840aad0b94bd6ece48c86065b24019895ff7fcf43a34e3a919R118-R120

Hmm `CancelBuilder::any()` (even without `.all()`) does seem to cancel every request in the ring, not just the first one found. 

AFAICT `CancelBuilder::any()` and `CancelBuilder::any().all()` are equivalent. Unless I'm missing something I don't think there is a way to say "cancel any single request".

gardnervickers: Maybe we can say "this will cancel any and all in-flight requests"?

FrankReh: My bad. If Any is already getting All, then I misunderstood, or mis-remembered, the man page. It never made sense to me that any would match just one, but I thought that's what they had implemented.

FrankReh (approved):
LGTM. Thanks!

src/types.rs
FrankReh: Never mind.

quininer:
LGTM. I will release 0.6.0 on wednesday.

Issue #208, opened 2023-03-01
Socket test fails on master


jordanisaacs:
I ran the `cargo run --package io-uring-test` and it consistently fails the socket test in a single configuration with an extra increment of the file descriptor. Other configurations pass:

```
ring type: IoUring<squeue::Entry, cqueue::Entry32>
params: Parameters {
    is_setup_sqpoll: false,
    is_setup_iopoll: false,
    is_setup_single_issuer: false,
    is_feature_single_mmap: true,
    is_feature_nodrop: true,
    is_feature_submit_stable: true,
    is_feature_rw_cur_pos: true,
    is_feature_cur_personality: true,
    is_feature_poll_32bits: true,
    sq_entries: 8,
    cq_entries: 16,
}
probe: {Op { code: 0 }, Op { code: 1 }, Op { code: 2 }, Op { code: 3 }, Op { code: 4 }, Op { code: 5 }, Op { code: 6 }, Op { code: 7 }, Op { code: 8 }, Op { code: 9 }, Op { code: 10 }, Op { code: 11 }, Op { code: 12 }, Op { code: 13 }, Op { code: 14 }, Op { code: 15 }, Op { code: 16 }, Op { code: 17 }, Op { code: 18 }, Op { code: 19 }, Op { code: 20 }, Op { code: 21 }, Op { code: 22 }, Op { code: 23 }, Op { code: 24 }, Op { code: 25 }, Op { code: 26 }, Op { code: 27 }, Op { code: 28 }, Op { code: 29 }, Op { code: 30 }, Op { code: 31 }, Op { code: 32 }, Op { code: 33 }, Op { code: 34 }, Op { code: 35 }, Op { code: 36 }, Op { code: 37 }, Op { code: 38 }, Op { code: 39 }, Op { code: 40 }, Op { code: 41 }, Op { code: 42 }, Op { code: 43 }, Op { code: 44 }, Op { code: 45 }, Op { code: 46 }, Op { code: 47 }, Op { code: 48 }}
...
test socket
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `11`', io-uring-test/src/tests/net.rs:1196:5
```

I am running stock linux kernel 6.2 and `rustc 1.67.1 (d5a82bbd2 2023-02-07)`

quininer:
This test seems require unstable kernel behavior. maybe we should remove this assert

RushilMehra:
[...]

PR #226, opened 2023-03-28
tests/net: relax FD assertion for socket opcode check

lucab:
This tweaks the socket opcode test in order to relax FD assertions. Checking for exact FD number can race when other threads are also opening/closing FDs, so just perform non-exact sanity checks instead.

Closes: https://github.com/tokio-rs/io-uring/issues/208


FrankReh:
And I thought this crate had to do the tests on a single thread. Turns out it was the tokio-uring crate were we limited it to a single thread for reasons like this. In that crate, we added a .cargo/config.toml file to set the test threads to 1.

PR #228, opened 2023-04-05
Added option to use SendZc for connectionless streams.

serzhiio:
With this changes we can set destination address. It is usefull when working with connectionless UDP, like `.send_to()`.
```
let std = SocketAddrV6::new(Ipv6Addr::new(1, 2, 3, 4, 5, 6, 7, 8), 9876, 11, 12);
let dest_addr = SockAddr::from(std);
let sendzc = opcode::SendZc::new(Fd(self.as_raw_fd()), buf.as_ptr(), buf.len() as _)
    .dest_addr(dest_addr.as_ptr())
    .dest_addr_len(dest_addr.len())
    .build()
    .user_data(1);
```

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -1,12 +1,12 @@
 use crate::utils;
 use crate::Test;
+use io_uring::types::Fd;
 use io_uring::{cqueue, opcode, squeue, types, IoUring};
 use once_cell::sync::OnceCell;
 use std::net::{TcpListener, TcpStream};
 use std::os::fd::FromRawFd;
 use std::os::unix::io::AsRawFd;
 use std::{io, mem};
-use io_uring::types::Fd;
 
 static TCP_LISTENER: OnceCell<TcpListener> = OnceCell::new();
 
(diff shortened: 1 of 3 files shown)


src/opcode.rs
```diff
@@ -1660,14 +1660,16 @@ opcode!(
         /// previously registered buffer. The buffer need not be aligned with the start of the
         /// registered buffer.
         buf_index: Option<u16> = None,
+        dest_addr: *const libc::sockaddr = core::ptr::null(),
```
quininer, line 1663: would you like to add some docs and tests?

serzhiio: Ouch... will try to :)

serzhiio:
test added

io-uring-test/src/tests/net.rs
```diff
@@ -1426,3 +1427,126 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+    // Check the completion events for the two UDP messages, plus a trailing
+    // CQE signaling that we ran out of buffers.
+    ring.submitter().submit_and_wait(3)?;
+    let cqes: Vec<cqueue::Entry> = ring.completion().map(Into::into).collect();
+
+    assert_eq!(cqes.len(), 7);
+
+    // first ZeroCopy notification for 33
+    assert_eq!(cqes[0].result(), 11);
+    assert_eq!(cqes[0].user_data(), 33);
```
quininer, line 1514: Is it too strict here? because I am not sure if kernel guarantees the order of cqes returns.

serzhiio:
With IO_LINK flag i hope we can rely on consistency

quininer:
Thank you!

PR #232, opened 2023-04-12
Connectionless stream send-recv test readability improvement

serzhiio:
Sending two data messages with destination address from client and receiving them on server stream with source address fetch.


quininer:
Thank you!

PR #236, opened 2023-06-27
fix alignment errors on i386

alexanderkjall:


mjt0k:
I think this whole testing for sizes/alignment is fundamentally wrong. Or we should know better about all the requirements of various platforms/architectures out there than bindgen.

quininer:
The code for `sys.rs` is auto-generated, do not change it manually. We can disable this test via options. https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.layout_tests

alexanderkjall:
Thanks, the context for this patch is that we are trying to package io-uring for debian, and i386 is one of our supported platforms: https://qa.debian.org/excuses.php?package=rust-io-uring

I guess this is really a bug in bindgen then, and that this one should be closed.

PR #237, opened 2023-06-29
Gate sc crate inclusion on the architectures it supports

alexanderkjall:
The sc crate is only functional on aarch64, riscv64 and x86_64.

This makes it only be included on those architectures, and fallback to the normal behavior on the others so that the programs still builds

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ io_safety = []
 bitflags = "1"
 
 libc = { version = "0.2.98", default-features = false }
+[target.'cfg(any(target_arch = "aarch64", target_arch = "riscv64", target_arch = "x86_64"))'.dependencies]
 sc = { version = "0.2", optional = true }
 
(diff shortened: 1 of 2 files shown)


quininer:
This doesn't make sense, if sc supports more arch in future, or users patch sc to support more arch, it will prevent it from working.

This is an optional feature, you just have to disable it on unsupported platforms.

alexanderkjall:
The context of this is the Debian package system, it verifies that it compiles and that the tests run successfully for each feature on all supported architectures and the direct-syscall didn't do that.  

So the currently solution is to disable it: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/io-uring/debian/patches/remove-optional-sc-dep.patch

I thought it would be better to be able to have it enabled for the supported architectures, and fall back to something else on the others. But I agree that it might introduce very confusing behavior and might be better to fail.

Thanks for the review.

PR #239, opened 2023-09-08
feat: create async cancel op with cancel builder

saiintbrisson:
Hi! [Kernel 5.19 introduced new flags](https://man7.org/linux/man-pages/man3/io_uring_prep_cancel.3.html) to the async cancel operation (that would
later also be used by `register_sync_cancel` in 6.0).

I hereby propose adding the ability to set up such flags within the existing `AsyncCancel`
operation, or via a new one.

I dislike having this not-so-obvious `builder` function on `AsyncCancel` but I also don't like
having a separate opcode solely to support the builder. Apart from the "hidden" builder feature,
using the same struct also mixes the new 5.19/6.0 (fixed FD) features.

Another option is to leave the flags to be manually set by the app, but why the extra hassle
if `CancelBuilder` is so handy? This was my first approach and the result didn't convince me.

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -79,6 +79,14 @@ fn test<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     tests::register_sync_cancel::test_register_sync_cancel(&mut ring, &test)?;
     tests::register_sync_cancel::test_register_sync_cancel_unsubmitted(&mut ring, &test)?;
     tests::register_sync_cancel::test_register_sync_cancel_any(&mut ring, &test)?;
+
+    // async cancellation
+    tests::cancel::test_async_cancel_user_data(&mut ring, &test)?;
+    tests::cancel::test_async_cancel_user_data_all(&mut ring, &test)?;
+    tests::cancel::test_async_cancel_any(&mut ring, &test)?;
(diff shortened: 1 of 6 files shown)


quininer:
This seems strange to me, because there are two places to set user_data. I would prefer to use a new opcode, and in next version we can delete old opcode.

saiintbrisson:
@quininer any name ideas? I've called it `AsyncCancelBuilder` before, but that might be misleading?

quininer:
I think AsyncCancel2 is good enough.

saiintbrisson:
@quininer Done. Should be ready for review.

quininer:
Can you add a unit test for this?

saiintbrisson:
@quininer Sure thing. Added unit tests for all variants and flags, except fixed FD.

saiintbrisson:
Completely unrelated question: could `Timespec` impl `From<Duration>`?

quininer:
Thank you!

Implementing From<Duration> seems reasonable.

saiintbrisson:
@quininer is there a release schedule for this crate? Is this change and the timespec one are eligible for a minor release?

quininer:
Sure, I can post a version.

quininer:
Sure, I can post a version.

PR #240, opened 2023-09-12
feat(types): timespec from duration

saiintbrisson:
Implements `From<Duration> for Timespec`. It just makes sense.


quininer:
Thank you!

PR #241, opened 2023-09-28
Add flags to async cancel

ollie-etl:
Adds the ability to set the flags to async cancel. Also makes public the existing `AsyncCancelFlags`

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -648,19 +648,20 @@ opcode! {
     pub struct AsyncCancel {
         user_data: { u64 }
         ;;
-
+        flags : types::AsyncCancelFlags = types::AsyncCancelFlags::empty()
         // TODO flags
     }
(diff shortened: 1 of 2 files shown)


src/opcode.rs
```diff
@@ -648,19 +648,20 @@ opcode! {
     pub struct AsyncCancel {
         user_data: { u64 }
         ;;
-
+        flags : types::AsyncCancelFlags = types::AsyncCancelFlags::empty()
         // TODO flags
```
ollie-etl, line 652:
```suggestion
        flags : types::AsyncCancelFlags = types::AsyncCancelFlags::empty()
```

ollie-etl:
@quininer

quininer:
see https://github.com/tokio-rs/io-uring/pull/239

PR #243, opened 2023-10-21
bug: fix overflow in `io-uring-test::register_buf_ring::InnerBufRing::buf_ring_push`

sw17ch:
With only c95a383de432dea6f73298dc8a998711d24cdc8f applied, running in debug mode results in a panic as the `old_tail + 1` logic will overflow when we've pushed more than `u16::MAX` times.

```
thread 'main' panicked at io-uring-test/src/tests/register_buf_ring.rs:308:24:
attempt to add with overflow
```

With 80b8982f290e287690b8c9e5d00668797a187557 applied, we properly wrap when we reach the boundary of `u16`.

### DIFF
--- a/io-uring-test/src/tests/register_buf_ring.rs
+++ b/io-uring-test/src/tests/register_buf_ring.rs
@@ -305,7 +305,7 @@ impl InnerBufRing {
         // for computing the ring entry, not to the tail value itself.
 
         let old_tail = self.local_tail.get();
-        self.local_tail.set(old_tail + 1);
+        self.local_tail.set(old_tail.wrapping_add(1));
         let ring_idx = old_tail & self.mask();
 
         let entries = self.ring_start.as_ptr_mut() as *mut BufRingEntry;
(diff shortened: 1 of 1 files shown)


io-uring-test/src/tests/register_buf_ring.rs
```diff
@@ -664,6 +664,15 @@ fn buf_ring_play<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     normal_check(&buf3, 1); // bid 1 should come back first.
     normal_check(&buf4, 0); // bid 0 should come back second.
 
+    std::mem::drop(buf3);
+    std::mem::drop(buf4);
+
+    // Now we loop u16::MAX+1 times to ensure proper behavior when the tail
+    // overflows the bounds of a u16.
+    for _ in 0..(u16::MAX as usize + 1) {
```
quininer, line 672: You can use `0..=u16::MAX`.

quininer (changes requested):

io-uring-test/src/tests/register_buf_ring.rs
sw17ch: Ah! Good point. I've fixed this in [a7aba2a](https://github.com/tokio-rs/io-uring/pull/243/commits/a7aba2a735a8bd72d5c85cab5060f1a8653d1294).

quininer:
Thank you!

PR #246, opened 2023-11-30
Add support for IORING_REGISTER_IOWQ_AFF

mjasny:
Hi,

I found that methods to specify cores for async workers are missing in this library.

Manpage: https://manpages.debian.org/unstable/liburing-dev/io_uring_register.2.en.html#IORING_REGISTER_IOWQ_AFF
Usage example in liburing: https://github.com/axboe/liburing/blob/cb00b6d807efaaaa5455223bf28a73fa216b2d24/test/wq-aff.c#L106

I decided to pull the libc::cpu_set_t related code into `register_iowq_aff`. If you think this is a bad idea, I'm happy to adapt the pull-request.

### DIFF
--- a/src/submit.rs
+++ b/src/submit.rs
@@ -399,26 +399,11 @@ impl<'a> Submitter<'a> {
     /// Tell io_uring on what CPUs the async workers can run. By default, async workers
     /// created by io_uring will inherit the CPU mask of its parent. This is usually
     /// all the CPUs in the system, unless the parent is being run with a limited set.
-    pub fn register_iowq_aff(&self, cores: &[u32]) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


quininer:
I think it would be better to just pass a `cpu_set_t`. can you add a test?

mjasny:
I modified the patch, such that the function accepts a `cpu_set_t`.

A test for setting the cpu mask of async workers is already in liburing itself (see second link above). However, if I would port this test to the `io-uring-test` package this would require a rewrite of the whole test-suite because a ring with sqpoll enabled is required. Please correct me if I'm wrong.

quininer:
You can create a new ring and write a simple test.

quininer:
Anyway, I'll merge it first. Thank you!

PR #248, opened 2023-12-21
pass `cqe->flags` with `MsgRingData`

bits0rcerer:
This PR allows to pass `cqe->flags` with `MsgRingData`.

based on: [axboe/liburing](https://github.com/axboe/liburing/blob/7524a6adf4d6720a47bfa617b5cb2fd8d57f16d2/src/include/liburing.h#L1032).

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1454,8 +1454,8 @@ opcode! {
         sqe.__bindgen_anon_1.off = user_data;
         sqe.__bindgen_anon_3.msg_ring_flags = opcode_flags;
         if let Some(_flags) = user_flags {
-            // TODO(lucab): add IORING_MSG_RING_FLAGS_PASS support (in v6.3):
-            // https://lore.kernel.org/all/20230103160507.617416-1-leitao@debian.org/t/#u
+            sqe.__bindgen_anon_5.file_index = _flags;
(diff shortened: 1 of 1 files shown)


src/opcode.rs
```diff
@@ -1454,8 +1454,8 @@ opcode! {
         sqe.__bindgen_anon_1.off = user_data;
         sqe.__bindgen_anon_3.msg_ring_flags = opcode_flags;
         if let Some(_flags) = user_flags {
```
quininer, line 1456: If it is used, the `_` prefix should be removed.

quininer (changes requested):

src/opcode.rs
```diff
@@ -1447,15 +1447,14 @@ opcode! {
         let MsgRingData { ring_fd, result, user_data, user_flags, opcode_flags } = self;
 
         let mut sqe = sqe_zeroed();
-        sqe.opcode = Self::CODE;
-        sqe.__bindgen_anon_2.addr = sys::IORING_MSG_DATA.into();
+        sqe.opcode = Self::CODE; sqe.__bindgen_anon_2.addr = sys::IORING_MSG_DATA.into();
```
quininer, line 1450: There seems to be a line break missing here.

bits0rcerer: Sorry. It seems like RustRover has some formatting problems in that file.
It removed all line breaks in my previous commit as well and I forgot to check this time.

quininer:
Thank you!

PR #250, opened 2024-01-03
Add various builder flags introduced in newer kernel versions

Dev380:
I added a few more flags from the [manpage.](https://man.archlinux.org/man/io_uring_setup.2.en) that seem like they're harmless to add to the API trivially.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,6 +301,15 @@ impl<S: squeue::EntryMarker, C: cqueue::EntryMarker> Builder<S, C> {
         self
     }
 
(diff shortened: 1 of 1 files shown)


quininer:
These are not builder flags, but feature flags filled by kernel. please refer to [Parameters::is_feature_single_mmap](https://docs.rs/io-uring/latest/io_uring/struct.Parameters.html#method.is_feature_single_mmap) to add it.

Dev380:
Sorry, I assumed everything that's configurable would be in Builder because monoio wants a Builder to create its runtime.

PR #254, opened 2024-01-21
Implement `AsFd` for all entry sizes

oxalica:
It seems to be a merge error of two concurrent PRs #133 and #134, which results in `AsFd` only implemented specifically on `IoUring<Entry, Entry>` rather than generically. This PR generalized that impl, also to align with `AsRawFd` impl.


quininer:
Thank you!

PR #256, opened 2024-01-26
Add safer abstractions for buffer rings

SUPERCILEX:
See https://github.com/tokio-rs/tokio-uring/issues/112#issuecomment-1903356303

I'm leaving this in the draft state until I've fully built out my server and fuzzed it while using this API to battle test it.

### DIFF
--- a/src/buf_ring.rs
+++ b/src/buf_ring.rs
@@ -1,0 +1,141 @@
+use std::mem::ManuallyDrop;
+use std::ops::{Deref, DerefMut};
+use std::{convert::TryFrom, marker::PhantomData, num::Wrapping, slice, sync::atomic};
+
+use crate::{sys, util::Mmap};
+
+pub struct BufRing {
+    ring: Mmap,
+    ring_entries: u16,
+    entry_size: u32,
+}
+
+impl BufRing {
+    pub(crate) fn init(ring: Mmap, ring_entries: u16, entry_size: u32) -> Self {
+        let mut this = Self {
+            ring,
+            ring_entries,
+            entry_size,
+        };
+
+        {
+            let mut s = this.submissions();
+            for i in 0u16..ring_entries {
+                let buf = unsafe { s._recycle_by_index(i) };
+                buf.len = entry_size;
+            }
+        }
+
+        this
+    }
+
+    pub fn submissions(&mut self) -> BufRingSubmissions<'_> {
+        let ring_ptr = self.ring.as_mut_ptr().cast::<sys::io_uring_buf>();
+        let tail_ptr = unsafe { self.ring.offset(8 + 4 + 2) };
+        let ring_entries = usize::from(self.ring_entries);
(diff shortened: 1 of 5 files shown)


src/buf_ring.rs
```diff
@@ -0,0 +1,87 @@
+use std::{convert::TryFrom, marker::PhantomData, num::Wrapping, slice, sync::atomic};
+
+use crate::{sys, util::Mmap};
+
+pub struct BufRing {
+    ring: Mmap,
+    ring_entries: u16,
+    entry_size: u32,
```
DylanZA, line 8: you might want to decouple this in the future.
Although its terribly complicated, you might want to reuse parts of existing buffers.

Lets say you have 1k buffers, but a message was received with only 8 bytes. In theory you could immediately requeue the rest of that buffer onto the buf_ring for the next network data. The hard part is obviously accounting for this and soon you might have built an allocator internally - but possibly worth allowing for this use case in the future.

SUPERCILEX: Yeah, originally I wanted to enable full flexibility, but that made the API a pain to use for the simple case. I forgot to mention that for creation I decided against allowing you to specify arbitrary buffer sizes because that effectively limits you to the smallest buffer size since the kernel can't pick the buffer size it wants and has to go in order. Similar logic applies for what you described: if the data is tiny and processing will take a while, you should probably just copy it off the buffer, but if the data is large, then you'll be adding back a small fraction of the buffer to the ring making it have limited utility when the kernel tries to use it.

Anyway, that was the rationale. But I don't think this current architecture prevents flexibility in the future (or I could add it now). I think you could have an unsafe get method get returns a buffer struct and an unsafe add method that adds a buffer struct. Then we could special case entry_size of 0 to not initialize any of the buffers so you can allocate your own.

DylanZA: i think that sounds generally sensible - was just in case you hadn't thought of it.

SUPERCILEX: Yeah, totally fair. :)

SUPERCILEX:
Ok, I've been running this for a while and I'm pretty confident it works at this point. I also expanded the API to support back pressure by carrying the buffer token through several io_uring steps before releasing it back to the kernel (that way if a client spams requests, it'll run out of buffers and need to wait until the server is ready to accept more).

I'm not really happy with the API, but I do think it gets the job done. Thoughts on it? I can add docs if we're going forward with it.

Also in case anyone's interested, here's how I'm using this: https://github.com/SUPERCILEX/clipboard-history/blob/418b2612f8e62693e42057029df78f6fbf49de3e/server/src/reactor.rs#L248

SUPERCILEX:
Closing as I'm probably the only one using this. The canonical implementation now lives in https://github.com/SUPERCILEX/clipboard-history/blob/9662f08b43502feafceec0c694237498bc7ea127/server/src/io_uring.rs#L167

PR #257, opened 2024-01-27
Fix broken RecvMsgOut parsing

SUPERCILEX:
The current parsing doesn't actually let the buffer ever be truncated and instead errors out immediately. This PR fixes it to only ever out when the buffer is smaller than the header. This implementation matches liburing's validation: https://github.com/axboe/liburing/blob/c44319a6bfb9953b430325227a94136690c19c33/src/include/liburing.h#L831-L834

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -1436,6 +1436,142 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
 
     Ok(())
 }
+pub fn test_udp_recvmsg_multishot_trunc<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> anyhow::Result<()> {
+    require!(
+        test;
+        test.probe.is_supported(opcode::RecvMsgMulti::CODE);
+        test.probe.is_supported(opcode::ProvideBuffers::CODE);
+        test.probe.is_supported(opcode::SendMsg::CODE);
+        test.check_kernel_version("6.6.0" /* 6.2 is totally broken and returns nonsense upon truncation */);
+    );
+
+    println!("test udp_recvmsg_multishot_trunc");
+
+    let server_socket: socket2::Socket = std::net::UdpSocket::bind("127.0.0.1:0").unwrap().into();
+    let server_addr = server_socket.local_addr()?;
+
+    const BUF_GROUP: u16 = 33;
+    const DATA: &[u8] = b"testfooo for me";
+    let mut buf1 = [0u8; 20]; // 20 = size_of::<io_uring_recvmsg_out>() + msghdr.msg_namelen
+    let mut buf2 = [0u8; 20 + DATA.len()];
+    let mut buf3 = [0u8; 20 + DATA.len()];
+    let mut buffers = [
+        buf1.as_mut_slice(),
+        buf2.as_mut_slice(),
+        buf3.as_mut_slice(),
+    ];
+
+    for (index, buf) in buffers.iter_mut().enumerate() {
+        let provide_bufs_e = io_uring::opcode::ProvideBuffers::new(
+            (**buf).as_mut_ptr(),
+            buf.len() as i32,
+            1,
+            BUF_GROUP,
+            index as u16,
+        )
+        .build()
+        .user_data(11)
+        .into();
+        unsafe { ring.submission().push(&provide_bufs_e)? };
+        ring.submitter().submit_and_wait(1)?;
+        let cqes: Vec<io_uring::cqueue::Entry> = ring.completion().map(Into::into).collect();
+        assert_eq!(cqes.len(), 1);
+        assert_eq!(cqes[0].user_data(), 11);
+        assert_eq!(cqes[0].result(), 0);
+        assert_eq!(cqes[0].flags(), 0);
+    }
+
+    // This structure is actually only used for input arguments to the kernel
+    // (and only name length and control length are actually relevant).
+    let mut msghdr: libc::msghdr = unsafe { std::mem::zeroed() };
+    msghdr.msg_namelen = 4;
+
+    let recvmsg_e = opcode::RecvMsgMulti::new(
+        Fd(server_socket.as_raw_fd()),
+        &msghdr as *const _,
+        BUF_GROUP,
+    )
+    .flags(libc::MSG_TRUNC as u32)
+    .build()
+    .user_data(77)
+    .into();
+    unsafe { ring.submission().push(&recvmsg_e)? };
+    ring.submitter().submit().unwrap();
+
+    let client_socket: socket2::Socket = std::net::UdpSocket::bind("127.0.0.1:0").unwrap().into();
+
+    let data = [io::IoSlice::new(DATA)];
+    let mut msghdr1: libc::msghdr = unsafe { mem::zeroed() };
+    msghdr1.msg_name = server_addr.as_ptr() as *const _ as *mut _;
+    msghdr1.msg_namelen = server_addr.len();
+    msghdr1.msg_iov = data.as_ptr() as *const _ as *mut _;
+    msghdr1.msg_iovlen = 1;
+
+    let send_msgs = (0..2)
+        .map(|_| {
+            opcode::SendMsg::new(Fd(client_socket.as_raw_fd()), &msghdr1 as *const _)
+                .build()
+                .user_data(55)
+                .into()
+        })
+        .collect::<Vec<_>>();
+    unsafe { ring.submission().push_multiple(&send_msgs)? };
+    ring.submitter().submit().unwrap();
+
+    ring.submitter().submit_and_wait(4).unwrap();
+    let cqes: Vec<io_uring::cqueue::Entry> = ring.completion().map(Into::into).collect();
+    assert_eq!(cqes.len(), 4);
+    let mut i = 0;
+    for cqe in cqes {
+        let is_more = io_uring::cqueue::more(cqe.flags());
+        match cqe.user_data() {
+            // send notifications
+            55 => {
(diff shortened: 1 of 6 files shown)


quininer:
This looks good, is there a regression test possible?

SUPERCILEX:
I added a test, but it's pretty terrible as I didn't want to deal with cancelling the recvmulti to change its header length or add ancillary data. That said, it does at least test for payload and namelen truncation.

SUPERCILEX:
@quininer any chance you can run the test on your machine? It works fine on mine, so I'm not sure what's going on in CI. Maybe old kernels didn't set the trunc flag?

quininer:
You can try check some flag to skip it on ci.

SUPERCILEX:
Sounds good, I added a kernel version check for 6.6 (which is what I'm running), but it might work on earlier kernels.

io-uring-test/src/tests/net.rs
```diff
@@ -1436,6 +1437,163 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+    require!(
+        test;
+        test.probe.is_supported(opcode::RecvMsgMulti::CODE);
+        test.probe.is_supported(opcode::ProvideBuffers::CODE);
+        test.probe.is_supported(opcode::SendMsg::CODE);
+    );
+
+    // Check kernel version
+    {
+        let mut uname: libc::utsname = unsafe { std::mem::zeroed() };
```
quininer, line 1453: This isn't the best solution in my opinion, but I can live with it.

Can you wrap it into a function like this?

```rust
fn check_kernel_version(version: &str) -> bool;
```

then put it into `require!` macro.

io-uring-test/src/tests/net.rs
```diff
@@ -1436,6 +1437,163 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+        unsafe {
+            assert!(libc::uname(&mut uname) >= 0);
+        }
+
+        match uname
+            .release
```
quininer, line 1459: we can introduce a crate to parse kernel version number.

quininer (changes requested):

io-uring-test/src/tests/net.rs
SUPERCILEX: > This isn't the best solution in my opinion

I agree, it's ugly lol.

---

Ok actually the new require macro isn't half bad.

io-uring-test/src/utils.rs
```diff
@@ -6,9 +6,17 @@ macro_rules! require {
         $test:expr;
         $( $cond:expr ; )*
     ) => {
-        let mut cond = true;
+        require!($test; min_kernel_version = 0,0; $( $cond; )*)
```
quininer, line 9: There's no need to modify the macros, I think it's fine to write like this

```rust
require!{
    test;
    check_kernel_version("6.6.0");
}
```

SUPERCILEX: Ah, I see. Done.

io-uring-test/src/tests/net.rs
SUPERCILEX: Done.

quininer:
maybe you are interested review this @lucab :)

io-uring-test/src/tests/net.rs
```diff
@@ -1436,6 +1436,135 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+
+    println!("test udp_recvmsg_multishot_trunc");
+
+    let server_socket: socket2::Socket = std::net::UdpSocket::bind("127.0.0.1:0").unwrap().into();
+    let server_addr = server_socket.local_addr()?;
+
+    const BUF_GROUP: u16 = 33;
+    const DATA: &[u8] = b"testfooo for me";
+    let mut buf1 = [0u8; 20];
+    let mut buf2 = [0u8; 20 + DATA.len()];
```
lucab, line 1459: The `20` here is a bit magic, it would be good to note where it is coming from (is it `size_of(io_uring_recvmsg_out)`?)

io-uring-test/src/tests/net.rs
```diff
@@ -1436,6 +1436,135 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+                assert!(cqe.result() > 0);
+                assert!(is_more);
+                let buf_id = io_uring::cqueue::buffer_select(cqe.flags()).unwrap();
+                let tmp_buf = &buffers[buf_id as usize];
+                let msg = types::RecvMsgOut::parse(tmp_buf, &msghdr);
+
+                match i {
+                    0 => {
+                        let msg = msg.unwrap();
+                        assert!(msg.is_payload_truncated());
```
lucab, line 1547: I think we can also check the incoming payload length here, and that `payload_data()` is an empty slice.

src/types.rs
```diff
@@ -407,45 +413,36 @@ impl<'buf> RecvMsgOut<'buf> {
                 .read_unaligned()
         };
 
-        let msghdr_name_len = msghdr.msg_namelen as _;
-        let msghdr_control_len = msghdr.msg_controllen as _;
-
-        // Check total length upfront, so that further logic here
-        // below can safely use unchecked/saturating math.
```
lucab, line 414: Would it be possible to bring back the saturating math in all the start/end computations below? I'm not sure why you changed them, but that doesn't seem to be part of the actual fix.

lucab (commented):
Thanks for the bugfix, looks generally correct to me.
I've left a few minor comments in review.

io-uring-test/src/tests/net.rs
```diff
@@ -1436,6 +1436,135 @@ pub fn test_udp_recvmsg_multishot<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+pub fn test_udp_recvmsg_multishot_trunc<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> anyhow::Result<()> {
+    require!(
+        test;
+        test.probe.is_supported(opcode::RecvMsgMulti::CODE);
+        test.probe.is_supported(opcode::ProvideBuffers::CODE);
+        test.probe.is_supported(opcode::SendMsg::CODE);
+        test.check_kernel_version("6.6.0");
```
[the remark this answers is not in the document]
SUPERCILEX: Oh, though actually I think you meant explain why I need 6.6? On that note, I think I can actually run this test on 6.2 and just put the 6.6 check for the flags. Lemme try. that.

src/types.rs
```diff
@@ -407,45 +413,36 @@ impl<'buf> RecvMsgOut<'buf> {
...
+                    msghdr_control_len,
+                );
+            let control_field_end = control_start + msghdr_control_len;
+            (&buffer[control_start..control_data_end], control_field_end)
         };
         let payload_data = {
-            let payload_data_end = payload_start.saturating_add(header.payloadlen as usize);
+            let payload_data_end = payload_start
+                + usize::min(
+                    usize::try_from(header.payloadlen).unwrap(),
```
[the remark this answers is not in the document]
SUPERCILEX: Seems reasonable, done.

src/types.rs
SUPERCILEX: They're unnecessary. We check the lengths above and use `min` everywhere, so it's impossible to overflow.

io-uring-test/src/tests/net.rs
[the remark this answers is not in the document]
SUPERCILEX: Nah I give up, the CI kernel is just super broken. Literally the only assert that passes in the first message is `assert!(msg.payload_data().is_empty());` and then `assert!(!msg.is_payload_truncated())` fails on the second message. I also tried flipping the message order around but that didn't work. I have no idea what's going on in the older kernel (and debugging by running CI is no fun)>

src/types.rs
SUPERCILEX: Sidenote: I filed https://github.com/rust-lang/rust/issues/120433 because we're not getting bounds check ellison which is annoying.

SUPERCILEX:
Ok, I think I addressed all the feedback.

src/types.rs
lucab: Yes we agree that they are all properly bound. That's why it seems reasonable to let the compiler know we aren't interested in panic points or overflowing math in there. Anyway, fine either ways.

io-uring-test/src/tests/net.rs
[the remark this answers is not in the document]
lucab: No worries, I was mostly suggesting if possible to anchor this to some kind of feature-probing (even if just a proxy-feature), as kernel version parsing tends to break on different distributions. I'm fine with keeping the version check at the moment.

lucab (approved):

quininer:
Thank you!

src/types.rs
SUPERCILEX: Oh I see, the saturating version does generate slightly better assembly with one less branch: https://rust.godbolt.org/z/Pdj3fxsnn. Through really I'd like to see LLVM fixed.

In the meantime, do you want a PR with saturating_add? Or we could also just tell LLVM stuff is in bound and then you get a completely branchless version:
```rust
if ..._end > buffer.len() {
  unsafe { std::hint::unreachable_unchecked(); }
}
```

io-uring-test/src/tests/net.rs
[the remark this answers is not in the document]
SUPERCILEX: Ah yeah, but there aren't any opcodes past 6.0 AFAIK.

src/types.rs
lucab: No need to, what we merged is also fine as is.

SUPERCILEX: Ok, thanks!

PR #260, opened 2024-02-08
Add fixed buffer support for uring_cmd

ollie-etl:
Adds a flag to uring_cmd16 and uring_cmd80 to support IORING_URING_CMD_FIXED flag. Available since kernel v6.1


quininer:
Sorry delay, Thank you!

PR #261, opened 2024-02-15
Add mutable RecvMsgOut

SUPERCILEX:
Parsing control data usually requires mutation to be safe (so parsing the same slice doesn't return multiple OwnedFds that point to the same FD).

I had to make the buffer fields public and use unsafe which is annoying, but since rust can't prove we don't alias the buffer it's the only way to get multiple slices out of it AFAIK.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -47,6 +47,7 @@ use std::convert::TryFrom;
 use std::marker::PhantomData;
 use std::num::NonZeroU32;
 use std::os::unix::io::RawFd;
+use std::slice;
 
 pub use sys::__kernel_rwf_t as RwFlags;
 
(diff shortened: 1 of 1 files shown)


quininer:
I don't really like this macro, is it possible to implement it using generics?

like

```rust
pub struct RecvMsgOut<BUF: MsgBuf> {
    // ...
}

trait MsgBuf {
    type Pointer;

    fn from_raw_parts(ptr: Self::Pointer, len: usize) -> Self;
    fn as_ptr(&self) -> Self::Pointer;
}

impl MsgBuf for &[u8] {
    type Pointer = *const u8;

    // ...
}

impl MsgBuf for &mut [u8] {
    type Pointer = *mut u8;

    // ...
}
```

SUPERCILEX:
Sorry, I'm a little unclear on what you're suggesting. What does the public API look like? Do we still have public fields with different mutability? Or are we keeping the public methods and implementing mutable variants when the generic parse input is a mut slice? And then does that mean use unsafe to create the slices rather than store them in the struct?

SUPERCILEX:
Closing as I'm probably the only one using this. The canonical implementation now lives in https://github.com/SUPERCILEX/clipboard-history/blob/9662f08b43502feafceec0c694237498bc7ea127/server/src/io_uring.rs#L28

PR #263, opened 2024-03-15
Correctly handle coop_taskrun

jrahman:
When coop taskrun or defer taskrun are set to defer running completion processing until the next system call (coop) or the start of the next io_uring_enter() call, the taskrun flag needs to be checked. The kernel will set the taskrun flag if there are deferred completions that need additional processing in kernel space before they're available to the application. Here we update the check in the submission path to ensure that flag is checked and GETEVENTS is passed into io_uring_enter() when needed

### DIFF
--- a/src/submit.rs
+++ b/src/submit.rs
@@ -63,10 +63,10 @@ impl<'a> Submitter<'a> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


quininer:
Can you provide a test?

PR #264, opened 2024-03-23
feat: allow uring builder to customize flags/features

bychjzh:
Currently, users cannot directly customize the flags and params of the uring builder, and with the rapid iteration of io_uring in the kernel, many new features will not be compatible with existing applications. So we need to provide more flexibility in the uring builder to allow customization of flags and params.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -407,12 +407,12 @@ impl<S: squeue::EntryMarker, C: cqueue::EntryMarker> Builder<S, C> {
         self
     }
 
-    pub fn with_flags(&mut self, flags : __u32) -> &mut Self {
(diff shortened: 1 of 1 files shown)


quininer:
I don't think this is as useful as you might think, since many `setup` are not enough to just change `.flags`, and it doesn't make sense to change `.features`.

bychjzh:
As an example, when I wanted to add the FEAT_FAST_POLL feature, there was no way I could add it to the original code.

quininer:
You may have misunderstood the `feature` field, it is not a parameter that you need to set, but you can use this field to check if the currently used kernel supports fast poll feature.

bychjzh:
Thanks for the heads up, have verified that I did misunderstand here.

PR #272, opened 2024-04-12
opcode: add Futex{Wake,Wait,Waitv}

oxalica:
These opcodes are available since Linux 6.7, and are mentioned by #253. `FutexWaitV` expects a new struct `futex_waitv` which is available in linux-raw-sys. An opaque type is added to the main crate to avoid polling in the dependency.

See:
- `IORING_OP_FUTEX_{WAIT,WAKE}`: https://github.com/torvalds/linux/commit/194bb58c6090e39bd7d9b9c888a079213628e1f6
- `IORING_OP_FUTEX_WAITV`: https://github.com/torvalds/linux/commit/8f350194d5cfd7016d4cd44e433df0faa4d4a703

### DIFF
--- a/io-uring-test/src/tests/futex.rs
+++ b/io-uring-test/src/tests/futex.rs
@@ -1,0 +1,182 @@
+use crate::Test;
+use io_uring::types::FutexWaitV;
+use io_uring::{cqueue, opcode, squeue, IoUring};
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use std::{io, ptr, thread};
+
+// Not defined by libc.
+//
+// From: https://github.com/torvalds/linux/blob/v6.7/include/uapi/linux/futex.h#L63
+const FUTEX2_SIZE_U32: u32 = 2;
+
+const USER_DATA: u64 = 0xDEAD_BEEF_DEAD_BEEF;
+const INIT_VAL: u32 = 0xDEAD_BEEF;
+
+fn syscall_futex(futex: *const u32, op: libc::c_int, val: u32) -> io::Result<i64> {
+    let ret = unsafe {
+        libc::syscall(
+            libc::SYS_futex,
+            futex,
+            op,
+            val,
+            ptr::null::<u8>(),
+            ptr::null::<u8>(),
+            0u32,
+        )
+    };
+    if ret >= 0 {
+        Ok(ret)
+    } else {
+        Err(io::Error::from_raw_os_error(-ret as _))
+    }
+}
+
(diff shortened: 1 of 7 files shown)


src/types.rs
```diff
@@ -64,6 +64,15 @@ pub struct epoll_event {
     _priv: (),
 }
 
+/// Opaque types, you should use [`futex_waitv`] instead.
+///
+/// [`futex_waitv`]: https://docs.rs/linux-raw-sys/latest/linux_raw_sys/general/struct.futex_waitv.html
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub struct futex_waitv {
```
quininer, line 72: For types missing in libc, I would prefer to export from sys mod, see `OpenHow` and `Timespec`.

quininer (changes requested):
Thanks. This looks pretty good!

src/types.rs
oxalica: > For types missing in libc, I would prefer to export from sys mod, see `OpenHow` and `Timespec`.

Do you mean only types? What about constants of `futex_flags`? Currently `FUTEX2_SIZE_U32` must be specified for these operations, or they will always fail.

quininer: I think just type is enough, user defining FUTEX2_SIZE_U32 seems risk-free. In the long term we will use libc's definitions.

oxalica:
Exposed `types::FutexWaitV` and also fixed a typo in tests (waitv, not wakev).

I noticed that io-uring-test fails `test_udp_recvmsg_multishot_trunc` on my machine (`io-uring-test/src/tests/net.rs:1549:25, assertion failed: msg.is_payload_truncated()`), not sure if it's something kernel related (I'm on 6.9.0-rc2). But at least it seems unrelated.

io-uring-test/src/tests/futex.rs
```diff
@@ -0,0 +1,182 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use std::{io, ptr, thread};
+
+// Not defined by libc.
+//
+// From: https://github.com/torvalds/linux/blob/v6.7/include/uapi/linux/futex.h#L63
+const FUTEX2_SIZE_U32: u32 = 2;
+
+const USER_DATA: u64 = 0xDEAD_BEEF_DEAD_BEEF;
```
quininer, line 14: I suggestion using different user_data for each entry to avoid an entry from one test is returned in next.

oxalica: I'm expecting the test harness construct individual io-uring for each test, but apparently it does not. :thinking:  Fixed now.

quininer (approved):
LGTM

quininer:
Thank you!

Issue #275, opened 2024-04-21
`ReadFixed` and `WriteFixed` descriptions


Connortsui20:
https://docs.rs/io-uring/latest/io_uring/opcode/struct.ReadFixed.html
> Read from pre-mapped buffers that have been previously registered...

Should this be "Read _into_ pre-mapped buffers" instead of "Read _from_"?

https://docs.rs/io-uring/latest/io_uring/opcode/struct.WriteFixed.html
> Write to pre-mapped buffers that have been previously registered...

Also, should this be "Write _from_ pre-mapped buffers" instead of "Write _to_"?

I'm pretty sure this is the case, but not 100% sure.

quininer:
You are right, would you like to open a PR to fix it?

Connortsui20:
Sure!

It would probably be good for me to take the time to add some more specific documentation that isn't just pointing to man pages. Hopefully I can get those done next month (when my school semester has finished)

PR #276, opened 2024-04-22
Update documentation for `ReadFixed` and `WriteFixed`

Connortsui20:
Adds some detail to the `ReadFixed` and `WriteFixed` opcodes (referred to in #275).

There is one small problem with this: I've documented the expected behavior according to the man pages, but I have actually yet to see these opcodes work in Rust. Are there any tests that check if these opcodes are correct? (My personal testing seems to not work, but it is very possible I've just set it up wrong since I haven't looked too deeply into it...)

Also, I am not completely sure why the `buf_index` is a `u16` type rather than a `i32`. It seems like the man pages for [`io_uring_prep_write_fixex`](https://man7.org/linux/man-pages/man3/io_uring_prep_write_fixed.3.html) allow for a 32-bit buffer index, and I don't think that there is a 16-bit limit on the number of buffers we are allowed to register (see the man page for [`io_uring_register_buffers`](https://man7.org/linux/man-pages/man3/io_uring_register_buffers.3.html), we are allowed to register an `unsigned` number of buffers).

I don't want to make this PR more than it should be, so I am happy to create another issue for the 2 things I mentioned.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -230,23 +230,11 @@ opcode! {
     /// Read from a file into a fixed buffer that has been previously registered with
     /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
     ///
-    /// The operation designated by this `opcode` will read data from a file at a specific offset
-    /// into a fixed buffer that has been registered into the `io-uring` instance via
-    /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
-    ///
-    /// The caller must ensure that the `buf_index` argument indexes into the list of buffers that
-    /// have been registered and that the buffer the index refers to has a base pointer of `buf` and
-    /// a length of `len`.
-    ///
     /// The return values match those documented in the `preadv2(2)` man pages.
     ///
     /// See the man pages for `io_uring_prep_read_fixed(3)` for more information.
     #[derive(Debug)]
     pub struct ReadFixed {
-        /// This operation is only valid if `buf_index` is an index into an array of previously
-        /// registered fixed buffers, and the `buf` and `len` arguments fall within the region
-        /// specified by `buf_index`. Note that `buf` does not need to be aligned with the start of
-        /// the registered buffer.
         fd: { impl sealed::UseFixed },
         buf: { *mut u8 },
         len: { u32 },
@@ -287,23 +275,11 @@ opcode! {
     /// Write to a file from a fixed buffer that have been previously registered with
     /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
     ///
-    /// The operation designated by this `opcode` will write data into a file at a specific offset
-    /// from a fixed buffer that has been registered into the `io-uring` instance via
-    /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
-    ///
-    /// The caller must ensure that the `buf_index` argument indexes into the list of buffers that
-    /// have been registered and that the buffer the index refers to has a base pointer of `buf` and
-    /// a length of `len`.
-    ///
     /// The return values match those documented in the `pwritev2(2)` man pages.
     ///
     /// See the man pages for `io_uring_prep_write_fixed(3)` for more information.
     #[derive(Debug)]
     pub struct WriteFixed {
-        /// This operation is only valid if `buf_index` is an index into an array of previously
-        /// registered fixed buffers, and the `buf` and `len` arguments fall within the region
-        /// specified by `buf_index`. Note that `buf` does not need to be aligned with the start of
-        /// the registered buffer.
         fd: { impl sealed::UseFixed },
         buf: { *const u8 },
         len: { u32 },
@@ -313,7 +289,7 @@ opcode! {
         /// The offset of the file to write to.
         offset: u64 = 0,
         /// Specified for write operations, contains a bitwise OR of per-I/O flags, as described in
-        /// the `preadv2(2)` man page.
+        /// the `pwritev2(2)` man page.
         rw_flags: types::RwFlags = 0
     }
 
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -231,8 +231,6 @@ opcode! {
     /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
     ///
     /// The return values match those documented in the `preadv2(2)` man pages.
-    ///
-    /// See the man pages for `io_uring_prep_read_fixed(3)` for more information.
     #[derive(Debug)]
     pub struct ReadFixed {
         fd: { impl sealed::UseFixed },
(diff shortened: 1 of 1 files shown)


src/opcode.rs
```diff
@@ -227,23 +227,36 @@ opcode! {
 }
 
 opcode! {
-    /// Read from pre-mapped buffers that have been previously registered with
+    /// Read from a file into a fixed buffer that has been previously registered with
     /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
     ///
+    /// The operation designated by this `opcode` will read data from a file at a specific offset
```
quininer, line 233: In my opinion there is no need to add too much description, something as simple as the manpage is fine.

src/opcode.rs
```diff
@@ -227,23 +227,36 @@ opcode! {
...
     /// The return values match those documented in the `preadv2(2)` man pages.
+    ///
+    /// See the man pages for `io_uring_prep_read_fixed(3)` for more information.
     #[derive(Debug)]
     pub struct ReadFixed {
-        /// The `buf_index` is an index into an array of fixed buffers,
-        /// and is only valid if fixed buffers were registered.
+        /// This operation is only valid if `buf_index` is an index into an array of previously
+        /// registered fixed buffers, and the `buf` and `len` arguments fall within the region
+        /// specified by `buf_index`. Note that `buf` does not need to be aligned with the start of
```
quininer, line 248: The added documentation seems unnecessary to me. Is this a behavior specific to ReadFixed?

quininer:
https://github.com/axboe/liburing/blob/liburing-2.5/src/include/liburing.h#L466
https://github.com/tokio-rs/io-uring/blob/master/src/sys/sys.rs#L743

`buf_index` is u16 because `buf_index` in entry can only be u16.

src/opcode.rs
```diff
@@ -227,23 +227,36 @@ opcode! {
...
+    /// into a fixed buffer that has been registered into the `io-uring` instance via
+    /// [`Submitter::register_buffers`](crate::Submitter::register_buffers).
+    ///
+    /// The caller must ensure that the `buf_index` argument indexes into the list of buffers that
+    /// have been registered and that the buffer the index refers to has a base pointer of `buf` and
+    /// a length of `len`.
+    ///
     /// The return values match those documented in the `preadv2(2)` man pages.
+    ///
+    /// See the man pages for `io_uring_prep_read_fixed(3)` for more information.
```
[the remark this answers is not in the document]
Connortsui20: Referencing the previous comment, do you know which man page the documentation should direct to? I understand that `liburing` is supposed to be a higher level thing that can change, but unless I missed something, there is no man page for the specific opcode `IORING_OP_READ_FIXED` and the write equivalent.

As a side note, regardless of if `liburing` does decide to change how `io_uring_prep_read_fixed` works, shouldn't this crate follow that? If `opcode` in this crate was just an enum like it exists in C I think that would make sense to keep it separate from `liburing`, but because in this crate you have to actually build an opcode struct to create a submission queue entry, I think synchronizing with `liburing` is reasonable (In fact, I thought this crate _was_ a stand-in for `liburing`, so that's on me).

src/opcode.rs
Connortsui20: Not completely sure what you mean by behavior "specific" to `ReadFixed`, it is behavior specific to both `io_uring_prep_xxxxx_fixed`: https://man7.org/linux/man-pages/man3/io_uring_prep_read_fixed.3.html

> The buf and nbytes arguments mus fall within a region specified by buf_index in the previously registered buffer. The buffer need not be aligned with the start of the registered buffer.

And from the first io_uring [article](https://kernel.dk/io_uring.pdf):

> Upon successful registration of the buffers, the application can use the IORING_OP_READ_FIXED and IORING_OP_WRITE_FIXED to perform IO to and from these buffers. When using these fixed op-codes, sqe->addr must contain an address that is within one of these buffers, and sqe->len must contain the length (in bytes) of the request.

src/opcode.rs
Connortsui20: The only reason I added so much documentation is because when I first tried to use this crate I had zero clue how to interpret this and I spent a long time trying to figure out how these opcodes were related to the io_uring C counterparts. However, if there is a better man page, that would certainly be better than this.

quininer:
I don't want docs to become too complex because I don't have the energy to track and maintain too much docs. It would be nice if you just fixed typos, or were willing to take on the maintenance of docs.

Connortsui20:
@quininer Honestly, I think I would be willing to take on maintenance of the docs. Though I have never done that before and also I have never actually used the raw C interface of `io_uring` and am only familiar with `liburing`, so out of caution I probably shouldn't take on such a large task that I'm not 100% familiar with.

As for this specific PR, if you are okay with just linking to the `liburing` page I can remove some extra docs.

FrankReh:
Yes, I always found the liburing man pages to be the best source for io_uring details. And the small group that manages them also is responsible for the io_uring kernel driver directly so any questions about their man pages and their header files are always quickly answered and they take PRs against their man pages very well. If the documentation here doesn't point to those man pages, it probably should.

Connortsui20:
@quininer I removed the extra docs so that now it just points to the `lib_uring` man pages, let me know if you think this is sufficient!

FrankReh:
@Connortsui20 I guess I didn't mean I thought it would help if individual opcode documentation in this repo pointed to the liburing man pages. Just that the whole usage of io-uring had to be understood by looking at the liburing man pages for details of the kernel API and then how the liburing headers worked off of that kernel API.

The liburing man pages don't try to describe all the details. They usually refer back to the original level 2 or level 3 manpage of the original kernel API.

Likewise this repo, I think, is taking the reasonable position that it would be too much to document details of each opcode. The user of this repo should just be educated somewhere at a high level for all its opcodes, that a clearer understanding can be gotten by reading through the liburing man pages and the liburing source, and sometimes the liburing unit test source.

The documentation here is a different level than that of liburing, but that's reasonable. It's remarkable that FaceTime affords the time to the authors to document and test as much as they do, for an open source project. Many projects, it seems, like having this repo for building against, and they undoubtedly dig into the details of how to make their use of the APIs work for them.

Connortsui20:
@FrankReh Yes that makes sense. However I don't think it's unreasonable for opcodes to point to specific man pages. Just because "most" people using this library should be educated on the high level usage does not mean this should purposefully exclude useful information. And additionally, educated people can forget where things are.

@quininer If you think that pointing to the man page is too much I'll remove it and open a new issue for discussion later.

quininer:
I agree with @FrankReh

Connortsui20:
@quininer Done

quininer:
Thank you!

Issue #277, opened 2024-04-23
`test_udp_recvmsg_multishot_trunc` fails


Connortsui20:
On my machine, this test fails. I'm running the most recent version of Arch Linux.

If I comment out each assertion one by one, every single assertion gets tripped.

Code is located [here](https://github.com/tokio-rs/io-uring/blob/501ee78049fa785eb4f5888252a0983ba6cf78e0/io-uring-test/src/tests/net.rs#L1547).

```rust
match i {
    0 => {
        let msg = msg.unwrap();
        assert!(msg.is_payload_truncated());
        assert!(msg.is_name_data_truncated());
        assert_eq!(DATA.len(), msg.incoming_payload_len() as usize);
        assert!(msg.payload_data().is_empty());
        assert!(4 < msg.incoming_name_len());
        assert_eq!(4, msg.name_data().len());
    }
```

Some potentially helpful error output:
```
assert_eq!(DATA.len(), msg.incoming_payload_len() as usize);

thread 'main' panicked at io-uring-test/src/tests/net.rs:1551:25:
assertion `left == right` failed
  left: 15
 right: 0
```

```
assert!(4 < msg.incoming_name_len(), "Name length: {}", msg.incoming_name_len());

thread 'main' panicked at io-uring-test/src/tests/net.rs:1553:25:
Name length: 0
```

```
assert_eq!(4, msg.name_data().len());

thread 'main' panicked at io-uring-test/src/tests/net.rs:1554:25:
assertion `left == right` failed
  left: 4
 right: 0
```

quininer:
@SUPERCILEX are you interested in checking this out? It changes in newer kernels.

SUPERCILEX:
What kernel version is this on? And do the `assert!(msg.is_payload_truncated())` assertions fail too? If not, then maybe the kernel just doesn't return anything anymore on truncation. If literally all the assertions fail, then I think it's time to open an io_uring bug because that would be nonsense: the data doesn't fit so it has day it's truncated.

Connortsui20:
@SUPERCILEX Sorry for the delay. Here is my linux version:
```
~
❯ uname -r
6.8.7-arch1-1

~
❯ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
[...]

PR #278, opened 2024-04-23
Test register buffers

Connortsui20:
Adds a test for [`register_buffers`](https://docs.rs/io-uring/latest/io_uring/struct.Submitter.html#method.register_buffers), [`WriteFixed`](https://docs.rs/io-uring/latest/io_uring/opcode/struct.WriteFixed.html), and [`ReadFixed`](https://docs.rs/io-uring/latest/io_uring/opcode/struct.ReadFixed.html).

This is branched off of #276.

There is a small technically "unnecessary" change related to #277 that just comments out the `tests::net::test_udp_recvmsg_multishot_trunc` test since it doesn't actually work on my machine. Feel free to revert that comment out!


quininer:
Thank you!

PR #279, opened 2024-04-24
feat: more derive traits for Fixed type

andrewgazelka:


quininer:
There's no harm in that, but `Fixed` is just an interface constraint and you're not expected to put it into a HashMap.

andrewgazelka:
> There's no harm in that, but `Fixed` is just an interface constraint and you're not expected to put it into a HashMap.

Is there any reason why putting it in a HashMap would be harmful? I think it can be nice to not have to convert across types.

quininer:
> There's no harm in that

There's no harm in that.

> Fixed is just an interface constraint

see https://github.com/tokio-rs/io-uring/blob/master/src/types.rs#L35

I can accept this PR, but I'm just reminding you that this is not how it is designed to be used. The type of `io_uring::types` should only be used when construct opcode.

andrewgazelka:
As this has not been merged in I am going to close this.

Issue #197, opened 2023-02-14
open question about race with kernel that liburing had to fix


FrankReh:
I found this commit to liburing some time ago. It may have no bearing on this repo because the atomic code is necessarily different anyway. But didn't want to lose track of this before I or someone else had made sure.

https://github.com/axboe/liburing/commit/744f4156b25d8630fee279cfd852ecc4c73952d4, 
- https://github.com/axboe/liburing/pull/542
fixing issue 
- https://github.com/axboe/liburing/issues/541

quininer:
I suspect that using `SeqCst` is not the most reasonable solution, but I don't mind following liburing

HeroicKatora:
My grains of salt, using SeqCst is a misinterpretation of the kernel documentation:

> * When using the SQ poll thread (IORING_SETUP_SQPOLL), the application
 * needs to check the SQ flags for IORING_SQ_NEED_WAKEUP *after*
 * updating the SQ tail; a full memory barrier smp_mb() is needed
 * between.

To the Rust memory model we need to establish the right happens-after relationship between the two operations. It should be sufficient to merely use `AcqRel` instead. Also, SeqCst semantically only affects the ordering with respect to _other_ SeqCst operations which is dubious since none of that is guaranteed by any of the interfaces.

HeroicKatora:
After having another think about this and re-re-reading the diagrams in the original issue, I don't think I really understood the cross-language-atomic-semantics properly yet. That might mean there are not the proper tools to properly solve this in Rust. Sorry for the noise.

A newer kernel which I was reading does use appear to be using an extra acquire fence. However it establishes another order than the one required for AcqRel to be definitely sufficient: https://github.com/torvalds/linux/blob/a93289b830ce783955b22fbe5d1274a464c05acf/io_uring/sqpoll.c#L334-L352

almogkh:
I had a look at the code here and the same bug is present here as in liburing/the kernel. The code here is almost the same as in liburing, the only difference is that liburing is using C/C++11 were as Rust follows the C++20 memory model but the differences between the two are not relevant to this bug.

> A newer kernel which I was reading does use appear to be using an extra acquire fence. However it establishes another order than the one required for AcqRel to be definitely sufficient

[...]

PR #281, opened 2024-04-27
Fix memory ordering in needs_wake

HeroicKatora:
A memory fence is required to ensure the decision of the kernel needing waking is consistent between user-space and kernel. The kernel must not go to sleep without us reliably observing the wake flag, respectively it must see updates to the submission head if our load does not contain updated wake flag. The fence is paired with one in the kernel which ensures that one of the two stores is observed by the respective load of the other party.

Fix: #197

### DIFF
--- a/src/squeue.rs
+++ b/src/squeue.rs
@@ -178,8 +178,43 @@ impl<E: EntryMarker> SubmissionQueue<'_, E> {
 
     /// When [`is_setup_sqpoll`](crate::Parameters::is_setup_sqpoll) is set, whether the kernel
     /// threads has gone to sleep and requires a system call to wake it up.
+    ///
+    /// A result of `false` is only meaningful if the function was called after the latest update
+    /// to the queue head. Other interpretations could lead to a race condition where the kernel
+    /// concurrently put the device to sleep and no further progress is made.
     #[inline]
     pub fn need_wakeup(&self) -> bool {
+        // See discussions that happened in [#197] and its linked threads in liburing. We need to
+        // ensure that writes to the head have been visible _to the kernel_ if this load results in
+        // decision to sleep. This is solved with a SeqCst fence. There is no common modified
+        // memory location that would provide alternative synchronization.
+        //
+        // The kernel, from its sequencing, first writes the wake flag, then performs a full
+        // barrier (`smp_mb`, or `smp_mb__after_atomic`), then reads the head. We assume that our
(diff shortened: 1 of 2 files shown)


almogkh:
Like I mentioned in the linked issue, I think the best place to put the `SeqCst` fence is in the `submit[_...]` functions right before calling `sq_need_wakeup` in submit.rs. I don't think the `after_intermittent_seqcst` function is necessary. We only need the full barrier when submitting new work to the SQ with SQPOLL enabled so executing the full barrier on submit makes the most sense.

HeroicKatora:
I do disagree with that, the library exposes access to the `SubmissionQueue` directly along with direct a raw `Submitter::enter`. If we're concerned with incorrect usage, this interface should itself already provide the sensible defaults instead of _only_ fixing the high-level usage through the library..

HeroicKatora:
I should put the fix in the right location though and add a barrier to the `Submitter::sq_need_wakeup` :weary:

almogkh:
The high-level interface in the `Submitter` doesn't call the `need_wakeup` function you modified, it uses its own function. I think at a minimum the safe high-level interface should behave correctly by default so regardless of the changes you made here, I think my suggested change should also probably be implemented.

I honestly don't know why `need_wakeup` is even exposed as a separate function in `SubmissionQueue`, I don't ever see a situation where you would want to query that separately from SQE submission. The equivalent function in liburing is not exposed. `Submitter::enter` is an unsafe function and the documentation for it says you should probably use `submit` instead. I think if you want to do all of this manually, it's fine to require users to insert the `SeqCst` fence manually, maybe with some documentation mentioning that it's necessary. I think forcing a `SeqCst` every time you want to check `need_wakeup` even if it's not needed isn't the best solution.

HeroicKatora:
Sure, the fix on `Submitter` can use separate barriers. There seems to be an existing open issue within the code about the duplicate read from `flags` that'll complicate the question further. If consolidated, the library definitely doesn't want to perform the barrier without sqpoll but code, maybe it's more straightforward to refactor.

I believe `SubmissionQueue` seems natural to me, the constraint is on the order of updating `head` with respect to reading the flag. As an external caller I do appreciate doing both through the same struct.

I'll ultimately leave this to maintainer's approach to the library interface. Commits are structure to include sufficient relevant documentation for each individual change (where it concerns public vs. private API).

quininer:
Thank you! Anyway I agree align to liburing.

PR #282, opened 2024-04-29
relax sq.flags atomic read

FrankReh:
Change the atomic read of the SQ flags field to be `relaxed`.

Refer to discussions in #197 and in particular, as pointed out in #197, refer to the liburing library loads of this same field.

The liburing library names this field sq.kflags and all its atomic loads are performed with their IO_URING_READ_ONCE macro, which it defines to be the `relaxed` atomic load.


quininer:
Thank you!

PR #283, opened 2024-04-30
unsafe impl `Send+Sync` for `Submitter`, `SubmissionQueue`, `CompletionQueue`

problame:
The `IoUring` type already has `Send` + `Sync` impls.

In [tokio-epoll-uring](https://github.com/neondatabase/tokio-epoll-uring), we use `split()` and need the `Submitter`, `SubmissionQueue` and `CompletionQueue` to be `Send` (not sync, though).

This PR adds those `unsafe impl`s, and adds hopefully not too repetitive commentary on why these impls are safe.

---------

Co-authored-by: Conrad Ludgate <conrad@neon.tech>
Cherry picked from commit https://github.com/neondatabase/fork--tokio-rs--io-uring/commit/bbc5a0c5f6cde9051037ec2fcc648cbadb7a80b4

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -168,6 +168,9 @@ fn test<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     // regression test
     tests::regression::test_issue154(&mut ring, &test)?;
 
+    // api tests
+    tests::api::test_sendness(&mut ring, &test)?;
+
     println!("Test count: {}", test.count.get());
 
     Ok(())
--- a/io-uring-test/src/tests/api.rs
+++ b/io-uring-test/src/tests/api.rs
@@ -1,0 +1,20 @@
+use crate::Test;
+use io_uring::IoUring;
+use io_uring::{cqueue, squeue};
+
+pub fn test_sendness<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    _test: &Test,
+) -> anyhow::Result<()> {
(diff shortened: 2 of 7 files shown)


quininer:
You seem to have to use unsafe to get `&'static mut` or use scope thread to make sense. I would recommend `submission_shared` and `completion_shared`.

problame:
Sorry, I'm not following. Who's talking about `&'static mut` ? 

> I would recommend submission_shared and completion_shared.

The unsafe impls proposed in this PR allow movement of Submitter, SubmissionQueue, and CompletionQueue, to different threads in consuming code, without any unsafe.

I consider that strictly better than the `unsafe fn submission_shared()` and `unsafe fn completion_shared()`.

Nehliin:
This would be useful in my project as well

DXist:
I think there is no one size fits all approach. Maybe a solution could be to enable Send via feature.

I'm personally biased towards the ring-per-thread setup recommended by the io_uring author. With non Send `IoUring` and `Submitter` interfaces it's possible to provide `push_shared` method since only one thread is allowed to enqueue io.

I prefer to queue all io operations and submit then in batches. During submission I might do random lookups in large in memory indices and expect frequent cache misses.

To hide cache miss latencies I could issue software prefetch instructions, do yield and interleave application logic using several coroutines.

Each coroutine could accept shared reference to `IO` driver that holds `IoUring` instance on Linux platforms. To allow each coroutine to submit new io operations there should be a `push`-like method that takes shared reference to thread-exclusive `Submitter` instance.

PR #285, opened 2024-05-11
Fix clippy and rustdoc warnings

Connortsui20:
Fixes a few rustdoc and clippy warnings.

The only thing that might not be good is my removal of the `try_from()` and `unwrap()`, but since it won't type check if the sys type changes I think it's fine.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -396,7 +396,7 @@ impl<'buf> RecvMsgOut<'buf> {
     #[allow(clippy::useless_conversion)]
     pub fn parse(buffer: &'buf [u8], msghdr: &libc::msghdr) -> Result<Self, ()> {
         let msghdr_name_len = usize::try_from(msghdr.msg_namelen).unwrap();
-        let msghdr_control_len = msghdr.msg_controllen;
+        let msghdr_control_len = usize::try_from(msghdr.msg_controllen).unwrap();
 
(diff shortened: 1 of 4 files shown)


src/types.rs
```diff
@@ -393,13 +393,14 @@ impl<'buf> RecvMsgOut<'buf> {
...
     #[allow(clippy::result_unit_err)]
+    #[allow(clippy::useless_conversion)]
     pub fn parse(buffer: &'buf [u8], msghdr: &libc::msghdr) -> Result<Self, ()> {
         let msghdr_name_len = usize::try_from(msghdr.msg_namelen).unwrap();
         let msghdr_control_len = usize::try_from(msghdr.msg_controllen).unwrap();
 
         if Self::DATA_START
             .checked_add(msghdr_name_len)
-            .and_then(|acc| acc.checked_add(msghdr_control_len))
+            .and_then(|acc| acc.checked_add(usize::try_from(msghdr.msg_controllen).unwrap()))
```
quininer, line 403: this seems unnecessary

Connortsui20: It's only to pass the CI

Connortsui20: Whoops forgot to change that back

quininer:
Thank you!

PR #287, opened 2024-05-22
Remove unused kernel diagnostic from io-uring-test

daniel-levin:
Hello there! I am a happy user of this library with a small contribution to make. I want to remove a field from the `Test` struct.

The `io-uring-test` crate is broken on any system for which `uname` returns a kernel version that doesn't conform to semver. This includes Fedora Linux, where `uname` might return a `release` such as `6.8.9-300.fc40.x86_64`. I was going to file an issue but then I noticed this code isn't actually used anywhere. By removing this field from `Test` I'd fix the issue for myself and others, while also removing the dependency on the `semver` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -346,7 +346,6 @@ dependencies = [
  "io-uring",
  "libc",
  "once_cell",
- "semver",
  "socket2",
  "tempfile",
 ]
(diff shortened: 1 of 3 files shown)


quininer:
Given that we're not using it, it's good to remove it. We can reintroduce it if necessary. Thank you!

PR #289, opened 2024-06-18
Fix nightly compiler complaints about size_of

SUPERCILEX:
Because the prelude now includes size_of, the nightly compiler really doesn't like it when you fully qualify size_of.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,9 +15,10 @@ mod sys;
 pub mod types;
 
 use std::marker::PhantomData;
+use std::mem::size_of;
 use std::mem::ManuallyDrop;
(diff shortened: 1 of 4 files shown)


SUPERCILEX:
I don't think the test failures are related BTW. It says `Error: Operation not permitted` which shouldn't have anything to do with this.

SUPERCILEX:
Yeah, I just checked by reverting the commit so the PR is blank and it's still broken. Very strange.

quininer:
I believe this is something that changed with github actions but I don't have the energy to deal with it.

SUPERCILEX:
Yeah fair enough. They might have blocked io_uring syscalls? Anyway, I'm fine leaving this here until someone finds a fix (don't have time to look into it either).

PR #290, opened 2024-07-01
feat: a convenient wrapper for buf ring

Berrysoft:
This is an implementation for a raw buffer ring that doesn't own the buffers, based on liburing impl. It will be convenient for developers to implement their own safe buffer ring.


quininer:
Sorry, due to work, I don't have energy to maintain more API at the moment. You can maintain it in your own project.

SUPERCILEX:
There's also https://github.com/tokio-rs/io-uring/pull/256

PR #294, opened 2024-07-21
Increase tail at once for push_multiple()

ileixe:
Tail should be increased atomically to be semantically same for link operation + sqpoll case.

We've found sqpoll_thread steal in the middle of linked ops which results in non in-order execution.


ileixe:
This was from my misunderstanding. The tail was a just cache.

PR #296, opened 2024-08-07
Update for recent kernels (up to 6.10)

yaroslavros:
Updated for new capabilities in kernels 6.8, 6.9, 6.10 including FixedFdInstall, FTruncate operations as well as bundled recv and send.
Added check for bundled send and recv feature flag (available from kernel 6.10).
Implemented tests accordingly.


yaroslavros:
Fixed formatting to keep rustfmt happy.

Not sure about "operation not permitted" in failing CI tests as they don't seem to indicate which specifically test failed - everything works fine on my machines. Is it a CI issue?
I see similar problem in few other PRs

quininer:
This is a ci issue, but we don't have manpower to fix it at the moment.

src/opcode.rs
```diff
@@ -1832,3 +1832,159 @@ opcode! {
...
+opcode! {
+    /// Receive a bundle of buffers from a socket.
+    ///
+    /// Parameter
+    ///     buf_group: The id of the provided buffer pool to use for the bundle.
+    ///
+    /// Note that as of kernel 6.10 first recv always gets a single buffer, while second
+    /// obtains the bundle of remaining buffers. This behavior may change in the future.
+    ///
+    /// Bundle variajnt is available since kernel 6.10
```
vmingchen, line 1926:
```suggestion
    /// Bundle variant is available since kernel 6.10
```

src/opcode.rs
```diff
@@ -1832,3 +1832,159 @@ opcode! {
...
+        Entry(sqe)
+    }
+}
+
+// === 6.9 ===
+
+opcode! {
+    /// Perform file truncation, equivalent to `ftruncate(2)`.
+    #[derive(Debug)]
+    pub struct FTruncate {
```
vmingchen, line 1869:
```suggestion
    pub struct Ftruncate {
```

To be consistent with `Fallocate` and `Fadvise`.

src/opcode.rs
```diff
@@ -1832,3 +1832,159 @@ opcode! {
...
+    pub fn build(self) -> Entry {
+        let RecvMultiBundle { fd, buf_group, flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
+        sqe.__bindgen_anon_4.buf_group = buf_group;
+        sqe.flags |= 1 << sys::IOSQE_BUFFER_SELECT_BIT;
+        sqe.ioprio = sys::IORING_RECV_MULTISHOT as _;
```
vmingchen, line 1986: Should this be
`
sqe.ioprio |= sys::IORING_RECV_MULTISHOT as u16
`?

vmingchen (commented):
Looks pretty good! It would be nice to have this in. I am looking for the `ftruncate` support.

I have a question about IORING_RECV_MULTISHOT and some nits.

vmingchen:
@yaroslavros Can you rebase and see if the CI issues are gone? Thanks!

src/opcode.rs
yaroslavros: I don't think so. Sqe starts empty and is set as a combination of IORING_RECV_MULTISHOT and IORING_RECVSEND_BUNDLE (following line)

src/opcode.rs
yaroslavros: Updated PR to resolve this

src/opcode.rs
yaroslavros: Updated PR to resolve this

yaroslavros:
@vmingchen I've synchronised my branch.

src/opcode.rs
```diff
@@ -1847,3 +1847,159 @@ opcode! {
...
+
+    pub const CODE = sys::IORING_OP_FIXED_FD_INSTALL;
+
+    pub fn build(self) -> Entry {
+        let FixedFdInstall { fd, file_flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        sqe.fd = fd.0 as _;
+        sqe.flags = 1 << sys::IOSQE_FIXED_FILE_BIT;
```
quininer, line 1873: I would prefer `Flags::FIXED_FILE.bits()`

src/opcode.rs
```diff
@@ -1847,3 +1847,159 @@ opcode! {
...
+    /// repeatedly posts a CQE when data is available. Each CQE will take a bundle of buffers
+    /// out of a provided buffer pool for receiving. The application should check the flags of each CQE,
+    /// regardless of its result. If a posted CQE does not have the IORING_CQE_F_MORE flag set then
+    /// the multishot receive will be done and the application should issue a new request.
+    ///
+    /// Note that as of kernel 6.10 first CQE always gets a single buffer, while second
+    /// obtains the bundle of remaining buffers. This behavior may change in the future.
+    ///
+    /// Multishot bundle variant is available since kernel 6.10.
+    pub struct RecvMultiBundle {
```
quininer, line 1983: I'm not sure if we want too many opcodes, it looks exactly like RecvMulti just with one more flag

quininer (changes requested):

src/opcode.rs
yaroslavros: Changed

src/opcode.rs
yaroslavros: I'm not sure what would be a better alternative in the current design. Exposing ioprio in API so that user could set IORING_RECVSEND_BUNDLE is quite messy as well as additional argument to existing calls that will break things.
Also keep in mind that RecvMultiBundle is available starting from kernel 6.10 while RecvMulti appeared back in 6.0.

It feels to me that having opcodes for those use cases is actually cleaner.

src/opcode.rs
```diff
@@ -1847,3 +1847,159 @@ opcode! {
...
+    pub fn build(self) -> Entry {
+        let SendBundle { fd, len, flags, buf_group } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.len = len;
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
+        sqe.ioprio |= sys::IORING_RECVSEND_BUNDLE as u16;
+        sqe.flags |= 1 << sys::IOSQE_BUFFER_SELECT_BIT;
```
quininer, line 1926: and this

src/opcode.rs
```diff
@@ -1847,3 +1847,159 @@ opcode! {
...
+
+    pub fn build(self) -> Entry {
+        let RecvBundle { fd, buf_group, flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
+        sqe.__bindgen_anon_4.buf_group = buf_group;
+        sqe.flags |= 1 << sys::IOSQE_BUFFER_SELECT_BIT;
```
quininer, line 1959: here

src/opcode.rs
```diff
@@ -1847,3 +1847,159 @@ opcode! {
...
+
+    pub fn build(self) -> Entry {
+        let RecvMultiBundle { fd, buf_group, flags } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
+        sqe.__bindgen_anon_4.buf_group = buf_group;
+        sqe.flags |= 1 << sys::IOSQE_BUFFER_SELECT_BIT;
```
quininer, line 2000: here

quininer (changes requested):

vmingchen (approved):
Thanks a lot!

src/opcode.rs
yaroslavros: Replaced

src/opcode.rs
yaroslavros: Replaced

src/opcode.rs
yaroslavros: Replaced

quininer:
Thank you!

vmingchen:
Hi @quininer , would you please release a new version now that this PR has added several new APIs. I am personally interested in the new `Ftruncate` function. Thanks!

PR #299, opened 2024-08-16
Add notif and related SendZc docs

erenon:
- **Add notif that checks for IORING_CQE_F_NOTIF**
- **Add extra documentation to SendZc**

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -1662,10 +1662,12 @@ pub fn test_udp_sendzc_with_dest<S: squeue::EntryMarker, C: cqueue::EntryMarker>
                 // First SendZc notification
                 11 => {
                     assert!(cqueue::more(cqe.flags()));
+                    assert!(!cqueue::notif(cqe.flags()));
                 }
                 // Last SendZc notification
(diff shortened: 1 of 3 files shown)


quininer:
Can you add a test? and you don't need to add much docs, just point to manpage of IORING_OP_SEND_ZC.

erenon:
Added a test. I have two issues with just linking the docs:

 - The quoted paragraph is in the middle of https://man7.org/linux/man-pages/man2/io_uring_enter.2.html , and I can't link that section directly (there's no anchor in the HTML)
 - Even if I link the doc, it wouldn't make the "notif" fn added by this PR discoverable

I find this doc+notif to be visible important, because without these, SendZc cannot be used correctly. If you feel strongly about keeping the doc minimal, I can of course replace the paragraph with a link, let me know.

quininer:
I don't have the energy to keep track of upstream docs, so I prefer to keep minimal description and point to correct upstream docs.

erenon:
Got it. Added a link instead of the text.

erenon:
fmt fixed (extra newline removed)

quininer:
Thank you!

PR #301, opened 2024-08-17
release 0.7.0

quininer:
* Update MSRV to 1.63
* Warning of potential ABI compatibility issues
* Update sys.rs to 6.10


quininer:
I'm a bit concerned that arch check will break too much project, so I plan to vendor more arch. this won't be done this weekend anyway (because black wukong).

PR #304, opened 2024-09-22
Fix name of some syscalls in docs

aconz2:


quininer:
I think you're right, but the docs comes from liburing manpage. https://github.com/axboe/liburing/blob/master/man/io_uring_enter.2#L1155

I would like docs to be simply consistent with manpage, would you like to fix liburing manpage as well?

aconz2:
Oh the plot thickens, makes sense, yes I will do that. Already confused about futex2 since that is almost a thing with `futex_{wait,wake,waitv` but there is no `futex2(2)` man page (nor for those calls) so I will try to figure that out and get back to you

aconz2:
updated to drop the futex2 changes, liburing now reflects the same changes

quininer:
Thank you!

PR #309, opened 2024-11-23
Implement the equivalent of `io_uring_prep_sendto`

ruuda:
In liburing, there is [`io_uring_prep_sendto`](https://github.com/axboe/liburing/blob/25b3d2d428bb6b374a4feea02ff488d41a7ba5a1/src/include/liburing.h#L873-L880), which is implemented as preparing a regular `send` followed by `set_addr` on the same SQE. This was not yet implemented here for `Send`, so add it.

Fortunately @serzhiio previously implemented the equivalent for `SendZc` in #228, so it was easy to adapt that for `Send`. Thanks serzhiio, this saved me a lot of time!

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1003,18 +1003,27 @@ opcode! {
         buf: { *const u8 },
         len: { u32 },
         ;;
-        flags: i32 = 0
+        flags: i32 = 0,
+
+        /// Set the destination address, for sending from an unconnected socket.
+        ///
+        /// When set, `dest_addr_len` must be set as well.
+        /// See also `man 3 io_uring_prep_send_set_addr`.
+        dest_addr: *const libc::sockaddr = core::ptr::null(),
+        dest_addr_len: libc::socklen_t = 0,
     }
 
     pub const CODE = sys::IORING_OP_SEND;
 
     pub fn build(self) -> Entry {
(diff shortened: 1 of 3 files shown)


src/opcode.rs
```diff
@@ -1003,18 +1003,27 @@ opcode! {
...
         ;;
-        flags: i32 = 0
+        flags: i32 = 0,
+
+        /// Set the destination address, for sending from an unconnected socket.
+        ///
+        /// When set, `dest_addr_len` must be set as well.
+        /// See also `man 3 io_uring_prep_send_set_addr`.
+        dest_addr: *const libc::sockaddr = core::ptr::null(),
+        dest_addr_len: libc::socklen_t = 0,
```
ruuda, line 1013: Because of the generated setters, this generates two of them. This is consistent with `SendZc`, but I think it would be nicer (for both, although it would be breaking to change `SendZc` now) to have a single `dest_addr(addr: *const libc::sockaddr, addr_len: socklen_t)`, so it is impossible to forget to specify the length. Is such a thing possible?

quininer:
Thank you!

src/opcode.rs
quininer: It's not too complicated so I think the status quo is good enough

ruuda:
Thanks for the quick release!

PR #310, opened 2024-12-05
Add support for ftruncate

vmingchen:
Introduces `opcode::Ftruncate` to support `ftruncate` which got supported since Linux 6.9.

Also updates the libc to the latest version to get the new sys::IORING_OP_FTRUNCATE flag from libc.

See

- `IORING_OP_FTRUNCATE` commit: https://github.com/torvalds/linux/commit/b4bb1900c12e6a0fe11ff51e1aa6eea19a4ad635
- liburing PR: https://github.com/axboe/liburing/pull/1004


vmingchen:
hmmm, just noticed https://github.com/tokio-rs/io-uring/pull/296 which include this already.

PR #313, opened 2025-01-22
lib: add `register_buffers*` functions

saiintbrisson:
Supersedes #217 and applies some changes to make names easier to recognize.

Based on @fathyb's work, the following register operations were added:
* `Submitter::register_buffers2` -> [`IORING_REGISTER_BUFFERS2`](https://manpages.debian.org/unstable/liburing-dev/io_uring_register.2.en.html#IORING_REGISTER_BUFFERS2),
* `Submitter::register_buffers_sparse` -> [`IORING_REGISTER_BUFFERS2`](https://manpages.debian.org/unstable/liburing-dev/io_uring_register.2.en.html#IORING_REGISTER_BUFFERS2) with the `IORING_RSRC_REGISTER_SPARSE`,
* `Submitter::register_buffers_update` -> [`IORING_REGISTER_BUFFERS_UPDATE`](https://manpages.debian.org/unstable/liburing-dev/io_uring_register.2.en.html#IORING_REGISTER_BUFFERS_UPDATE).
  * If `tags = None`, tags will be disabled.

The `register_buffers2` has this name to better represent the opcode, and this is a common thing in this codebase (e.g. `AsyncCancel2`).


src/lib.rs
saiintbrisson, line 1: Clippy complained 😅

io-uring-test/src/tests/register_buffers.rs
saiintbrisson, line 1: The changes were actually minimal but git-diff doesn't think the same way as I do.

quininer: If you put test_register outside of test_register_buffers , git-diff might be happy.

saiintbrisson: that's right, it helped, thanks!

quininer (approved):
LGTM

PR #314, opened 2025-02-16
Fix build with bindgen feature

beviu:
Since a2fed13cee293ac3cf5bef51c4b0d5fc5a2a5a69, when the bindgen feature is enabled, the build script writes the bindings to `OUT_DIR/sys_<arch>.rs` (or `src/sys/sys_<arch>.rs` when the overwrite feature is enabled). However, the `src/sys/mod.rs` file still imports the file from `OUT_DIR/sys.rs` (without the architecture in the file name).

`OUT_DIR` already depends on the architecture, so there is no need to include the architecture in the file name.

Modify the build script to write the file to `OUT_DIR/sys.rs` so that it can be included by `src/sys/mod.rs`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -21,14 +21,6 @@ fn build() {
 #include <linux/futex.h>
     "#;
 
-    #[cfg(not(feature = "overwrite"))]
-    let outdir = PathBuf::from(env::var("OUT_DIR").unwrap());
-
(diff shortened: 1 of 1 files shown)


quininer:
Thank you!

jackpot51:
@quininer can you publish a new release for this?

PR #316, opened 2025-02-16
Allow more than two receives in test_tcp_recv_bundle

beviu:
On my machine with Linux 6.12, 4 receives were needed to drain the socket but the test assumes that only 2 are needed. The comment also said this could change in the future.


quininer:
That makes sense. Thank you.

PR #317, opened 2025-02-23
feat: add support for ReadMulti opcode (6.7+)

kp-omer-shamash:
Simple addition, already have the CODE in the /sys/

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -901,6 +901,47 @@ opcode! {
 }
 
 opcode! {
+    /// Read multiple times from a file, equivalent to `pread(2)`.
+    ///
+    /// Parameter:
+    ///     buf_group: The id of the provided buffer pool to use for each received chunk.
+    ///
+    /// MSG_WAITALL should not be set in flags.
+    ///
+    /// The multishot version allows the application to issue a single read request, which
+    /// repeatedly posts a CQE when data is available. Each CQE will take a buffer out of a
+    /// provided buffer pool for receiving. The application should check the flags of each CQE,
+    /// regardless of its result. If a posted CQE does not have the IORING_CQE_F_MORE flag set then
+    /// the multishot read will be done and the application should issue a new request.
+    ///
+    /// Multishot read is available since kernel 6.7.
+    /// Multishot read is suggested since kernel 6.7.2, see: https://github.com/axboe/liburing/issues/1041
+
+    pub struct ReadMulti {
+        fd: { impl sealed::UseFixed },
+        buf_group: { u16 },
+        ;;
+        ioprio: u16 = 0,
+        flags: i32 = 0
+    }
+
+    pub const CODE = sys::IORING_OP_READ_MULTISHOT;
+
+    pub fn build(self) -> Entry {
+        let ReadMulti { fd, buf_group, flags, ioprio } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
+        sqe.__bindgen_anon_4.buf_group = buf_group;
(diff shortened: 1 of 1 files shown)


kp-omer-shamash:
Right now I'm doing what feels like a crime against humanity by implementing like this and having tons of structures copy-pasta'd from the /sys/ part of the crate, would really appreciate if we can merge so I don't use this abomination :
```
pub struct ReadMulti {
    fd: i32,
    buf_group: u16,
    flags: i32,
}

impl ReadMulti {
    #[inline]
    pub fn new(fd: i32, buf_group: u16) -> Self {
        ReadMulti {
            fd,
            buf_group,
            flags: 0,
        }
    }

    #[inline]
    pub fn build(self) -> Entry {
        let sqe = CustomSQE {
            opcode: IORING_OP_READ_MULTISHOT as _,
            flags: io_uring::squeue::Flags::BUFFER_SELECT.bits(),
            fd: self.fd,
            buf_index: PackedU16 {
                buf_index: self.buf_group,
            },
            msg_flags: Union3 {
                msg_flags: self.flags as _,
            },
            ..Default::default()
        };

        // Safety: CustomSQE has identical memory layout to io_uring_sqe
        #[allow(unsafe_code)]
        unsafe {
            std::mem::transmute(sqe)
        }
    }
}
```

quininer:
This looks good, would you like to add a test?

kp-omer-shamash:
@quininer done

src/opcode.rs
```diff
@@ -900,6 +900,47 @@ opcode! {
...
+    /// provided buffer pool for receiving. The application should check the flags of each CQE,
+    /// regardless of its result. If a posted CQE does not have the IORING_CQE_F_MORE flag set then
+    /// the multishot read will be done and the application should issue a new request.
+    ///
+    /// Multishot read is available since kernel 6.7.
+    /// Multishot read is suggested since kernel 6.7.2, see: https://github.com/axboe/liburing/issues/1041
+
+    pub struct ReadMulti {
+        fd: { impl sealed::UseFixed },
+        buf_group: { u16 },
```
quininer, line 922: Do we need to provide a way to specify len and offset?

src/opcode.rs
```diff
@@ -900,6 +900,47 @@ opcode! {
...
+
+    pub const CODE = sys::IORING_OP_READ_MULTISHOT;
+
+    pub fn build(self) -> Entry {
+        let ReadMulti { fd, buf_group, flags, ioprio } = self;
+
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_3.msg_flags = flags as _;
```
quininer, line 936: What does this `flags` mean and why does it write to `msg_flags`?

src/opcode.rs
kp-omer-shamash: No, it should act the exact same as `RecvMulti` - you cannot specify anything other then which buffer-group to use, fd and flags (which I think none are not used at kernel-side right now)

src/opcode.rs
kp-omer-shamash: same as above - the API is similar to `RecvMulti`
flags are here in case kernel maintainers support any custom flags and we would like to use them, but default to 0

kp-omer-shamash (commented):
Sorry for late reply, I'm currently no longer available in this account, I hope this can be merged

src/opcode.rs
quininer: I think the kernel doesn't read `msg_flags`, right?

src/opcode.rs
quininer: I think this is not true, ReadMulti is more similar to Read than to RecvMulti.

see https://github.com/torvalds/linux/blob/v6.13/io_uring/rw.c#L366 and https://github.com/axboe/liburing/blob/master/src/include/liburing.h#L872

Issue #293, opened 2024-07-11
Support io_uring_prep_cmd_sock


erenon:
Hi, thanks for the nice library! Is [io_uring_prep_cmd_sock](https://man7.org/linux/man-pages/man3/io_uring_prep_cmd.3.html) supported? If not, is there a fundamental reason it is not supported?

Use case: I'm using io_uring `AcceptMulti` with `allocate_file_index`, and I register files with `register_files_sparse`. Therefore I have `FixedFd`s only, so I can't use the regular sync setsockopt/getsockopt interface, lacking a normal fd.

erenon:
Apparently, this is implemented via [UringCmd16](https://docs.rs/io-uring/latest/io_uring/opcode/struct.UringCmd16.html)/UringCmd80.

ThomasHabets:
@erenon could you elaborate / give an example? I also need to call `setsockopt()` with io_uring.

erenon:
https://man7.org/linux/man-pages/man3/io_uring_prep_cmd.3.html

Does this help?

ThomasHabets:
Thanks for the link. I'd already found it, but not found it directly helpful. It's for the C API.

Or are you saying as a starting point for creating a PR adding `setsockopt` support to `io-uring`?

erenon:
I think it is doable with the current API, but ugly indeed. Here's the C liburing impl to create the right SQE:

https://github.com/axboe/liburing/blob/0564cda79e91345bf5e6180e33de23a01e4aadde/src/include/liburing.h#L1328

```c
IOURINGINLINE void io_uring_prep_cmd_sock(struct io_uring_sqe *sqe,
					  int cmd_op,
					  int fd,
					  int level,
					  int optname,
					  void *optval,
					  int optlen)
{
	io_uring_prep_rw(IORING_OP_URING_CMD, sqe, fd, NULL, 0, 0);
	sqe->optval = (unsigned long) (uintptr_t) optval;
	sqe->optname = optname;
	sqe->optlen = optlen;
	sqe->cmd_op = cmd_op;
	sqe->level = level;
}
```

And here's the rust code that populates the SQE, created by UringCmd16:
https://docs.rs/io-uring/latest/src/io_uring/opcode.rs.html#1511

With some fiddling, you should be able to arrange the bytes in `cmd`, to fill the fields correctly (level, optname, optval, optlen)

Possibly you need UringCmd80? Ugly, that's for sure.

ThomasHabets:
[...]

PR #320, opened 2025-04-05
Add setsockopt

ThomasHabets:
See discussion at https://github.com/tokio-rs/io-uring/issues/293


quininer:
Can you add a test?

ThomasHabets:
Not entirely clear where the best place would be to put it, but I've now added a second commit with a test.

src/opcode.rs
```diff
@@ -641,6 +641,36 @@ opcode! {
...
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_1.__bindgen_anon_1.cmd_op = sys::SOCKET_URING_OP_SETSOCKOPT;
+
+        sqe.__bindgen_anon_2.__bindgen_anon_1.level = level;
+        sqe.__bindgen_anon_2.__bindgen_anon_1.optname = optname;
+        sqe.__bindgen_anon_3.uring_cmd_flags = flags;
+        sqe.__bindgen_anon_5.optlen = optlen;
+        sqe.__bindgen_anon_6.bindgen_union_field[0] = optval as _;
```
quininer, line 669: I think it's better to write via `.optval` field.

quininer (changes requested):

src/opcode.rs
```diff
@@ -641,6 +641,36 @@ opcode! {
...
+    }
+
+    pub const CODE = sys::IORING_OP_URING_CMD;
+
+    pub fn build(self) -> Entry {
+        let SetSockOpt { fd, level, optname, optval, optlen, flags } = self;
+        let mut sqe = sqe_zeroed();
+        sqe.opcode = Self::CODE;
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_1.__bindgen_anon_1.cmd_op = sys::SOCKET_URING_OP_SETSOCKOPT;
```
quininer, line 663: Should we expose `cmd_op`? Provide generalized `CmdSock` opcode instead of specific `SetSockOpt`.

`SOCKET_URING_OP_SETSOCKOPT` can be exposed as a `SocketOp` in `types.rs`.
like

```rust
pub struct SocketOp(u32);

impl SocketOp {
    pub const SETSOCKOPT: SocketOp = SocketOp(sys::SOCKET_URING_OP_SETSOCKOPT);
}
```

quininer (changes requested):

src/opcode.rs
ThomasHabets: Not sure what you mean. `level`, `optname`, `optval`, and `optlen` are all `setsockopt` specific, so we'd want an opcode creator either way.

Non-`setsockopt` needs other parameters.

Sure, it'd be possible to add an abstraction layer inside this crate to deduplicate, but honestly the duplicate code is like one single simple assignment line.

Maybe I'm not understanding your suggestion?

src/opcode.rs
ThomasHabets: Requires `unsafe`, but probably yes. Updated.

ThomasHabets:
Hang on, let me test this a bit.

src/opcode.rs
quininer: Ok, in any case there are no other needs at moment. new opcode can be added in future if needed.

quininer:
Thank you!

PR #324, opened 2025-04-29
vendor riscv64 sys.rs

Xeonacid:
Tested on a riscv64 device with io-uring-test.

If applicable, please make a new release for this so downstream can make use of this,  as we Arch Linux RISC-V are facing a ton of build errors for packages. We need to patch them one by one to add the `bindgen` feature.
```rust
error: The prebuilt `sys.rs` may not be compatible with your target,
       please use bindgen feature to generate new `sys.rs` of your arch
       or use `--cfg=io_uring_skip_arch_check` to skip the check.
  --> /build/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/io-uring-0.7.4/src/sys/mod.rs:23:1
   |
23 | / compile_error!(
24 | |     "The prebuilt `sys.rs` may not be compatible with your target,
25 | | please use bindgen feature to generate new `sys.rs` of your arch
26 | | or use `--cfg=io_uring_skip_arch_check` to skip the check."
27 | | );
   | |_^
```
Thanks!


quininer:
Thank you.

before a new version is released, you can use `--cfg=io_uring_use_own_sys` and `IO_URING_OWN_SYS_BINDING=$FILE` to point to your own binding files.

src/sys/sys_riscv64.rs
```diff
@@ -0,0 +1,3404 @@
+/* automatically generated by rust-bindgen 0.69.4 */
```
quininer, line 1: This `sys.rs` lacks `IORING_FEAT_RW_ATTR` compared to other arch, what the version of header file does it come from?

Xeonacid: Oops. Now built with 6.14 linux-api-headers, and the constant appears. Thanks for the review!

quininer:
Thank you!

PR #325, opened 2025-05-01
feat: Support Bind and Listen opcodes

### LINKED ISSUE #318: Add support for bind/listen
https://github.com/axboe/liburing/wiki/What%27s-new-with-io_uring-in-6.11-and-6.12
> Added in 6.11. See the liburing io_uring_prep_bind(3) and io_uring_prep_listen(3) man pages for details.

[bind](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=7481fd93fa0a851740e26026485f56a1305454ce), [listen](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ff140cc8628abfb1755691d16cfa8788d8820ef7)

pchaseh:
Closes #318

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -1469,20 +1469,19 @@ pub fn test_socket<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     require!(
         test;
         test.probe.is_supported(opcode::Socket::CODE);
-        test.probe.is_supported(opcode::Bind::CODE);
     );
 
     println!("test socket");
 
     // Open a UDP socket, through old-style `socket(2)` syscall.
     // This is used both as a kernel sanity check, and for comparing the returned io-uring FD.
-    let plain_socket = Socket::new(Domain::IPV4, Type::STREAM, Some(Protocol::TCP)).unwrap();
+    let plain_socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP)).unwrap();
     let plain_fd = plain_socket.as_raw_fd();
 
     let socket_fd_op = opcode::Socket::new(
         Domain::IPV4.into(),
-        Type::STREAM.into(),
-        Protocol::TCP.into(),
+        Type::DGRAM.into(),
+        Protocol::UDP.into(),
     );
     unsafe {
         ring.submission()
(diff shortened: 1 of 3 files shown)


io-uring-test/src/tests/net.rs
```diff
@@ -1469,19 +1469,20 @@ pub fn test_socket<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     require!(
         test;
         test.probe.is_supported(opcode::Socket::CODE);
+        test.probe.is_supported(opcode::Bind::CODE);
     );
 
     println!("test socket");
 
     // Open a UDP socket, through old-style `socket(2)` syscall.
```
quininer, line 1477: This changes the behavior mentioned in comment

src/opcode.rs
```diff
@@ -2033,3 +2033,49 @@ opcode! {
...
+        assign_fd!(sqe.fd = fd);
+        sqe.__bindgen_anon_2.addr = addr as _;
+        sqe.__bindgen_anon_1.off = addrlen as _;
+        Entry(sqe)
+    }
+}
+
+opcode! {
+    /// Listen on a socket, equivalent to `listen(2)`.
+    pub struct Listen {
```
quininer, line 2064: better add a test for it too

quininer (changes requested):

quininer:
Sorry delay, I missed this.

io-uring-test/src/tests/net.rs
pchaseh: I'm assuming you meant the comment about UDP sockets when I've changed the test to use TCP sockets. I've split out these changes to have their own test with a corrected comment. This way, there's still some functioning coverage for sockets even if the tests aren't being ran on kernel >= 6.11.

src/opcode.rs
pchaseh: The `Listen` opcode should already be covered here, but if you think it needs to be tested in some other way do let me know please
https://github.com/tokio-rs/io-uring/pull/325/files#diff-f038c03cda731ae8f6159ec51c7db83e68d1b036c046243c1c89939a72b3afd9R1668

quininer (approved):
LGTM

quininer:
Thank you!

PR #328, opened 2025-05-05
Add loongarch64 sys.rs

OriginCode:
Tested with `io-uring-test`.


quininer:
Do you have any actual needs? Since I don't have a loongarch machine, I can't really test and update it.

OriginCode:
> Do you have any actual needs? Since I don't have a loongarch machine, I can't really test and update it.

Yes, the distribution that I maintain supports loongarch64 as tier 1 architecture, therefore we need this.

quininer:
These PRs seem to have been generated from a more backward version of kernel headers?

PR #329, opened 2025-05-06
Use libc instead of hardcoded errnos

OriginCode:
Fix tests for weird architectures like mips64.

### DIFF
--- a/io-uring-test/src/tests/net.rs
+++ b/io-uring-test/src/tests/net.rs
@@ -687,7 +687,7 @@ pub fn test_tcp_accept_multi<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     assert_eq!(cqes[op1].user_data(), 2002);
     assert_eq!(cqes[op2].user_data(), 2003);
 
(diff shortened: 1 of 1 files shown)


io-uring-test/src/tests/net.rs
```diff
@@ -687,7 +687,7 @@ pub fn test_tcp_accept_multi<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     assert_eq!(cqes[op1].user_data(), 2002);
     assert_eq!(cqes[op2].user_data(), 2003);
 
-    assert_eq!(cqes[op1].result(), -125); // -ECANCELED
+    assert_eq!(cqes[op1].result(), -libc::ECANCELED); // -ECANCELED
```
bryantbiggs, line 690: The trailing doc comment is a bit redundant now - could be removed

OriginCode: Addressed

bryantbiggs (approved):
lgtm - will need someone with approval rights though

quininer:
Thank you!

PR #332, opened 2025-05-09
feat: Support `IORING_TIMEOUT_MULTISHOT` flag

### LINKED ISSUE #331: Multishot timeout is unsupported
See https://man7.org/linux/man-pages/man3/io_uring_prep_timeout.3.html, we're just missing exposing `sys::IORING_TIMEOUT_MULTISHOT` as part of [TimeoutFlags](https://github.com/tokio-rs/io-uring/blob/master/src/types.rs#L92)

pchaseh:
I've also removed the comment for `flags` which was already non-exhaustive (meaning doesn't cover `ETIME_SUCCESS` , `BOOTTIME`, or `REALTIME`). I think it's better to leave the description of the possible flags in `TimeoutFlags` instead so as to avoid duplicating this information to every field that uses said type

Closes #331


quininer:
Thank you!

PR #335, opened 2025-05-19
vendor loongarch64 sys.rs

k0tran:
Hi! I'm trying to build cosmic DE on loongarch64.
One of components - cosmic-files depends on compio which depends on io-uring.
As I'm aware there is no way for cargo to specify bindgen feature for dependency of a dependency (io-uring).
Because of that I generated required sys.rs and did some dirty patching.
It seems to work but I would prefer if sys_loongarch64.rs was already vendored. Would you mind that?

Tested it with io-uring-test and cargo test


quininer:
For now I recommend specifying your own `sys.rs` using `--cfg=io_uring_use_own_sys` + `IO_URING_OWN_SYS_BINDING=$FILE`.

quininer:
Closed because https://github.com/tokio-rs/io-uring/pull/328 was merged

PR #338, opened 2025-05-31
Add missing `register_files` APIs

ChillFish8:
This PR adds two missing ring calls, `io_uring_register_files_tags` and `io_uring_register_files_update_tag`. 

Tests have been added and confirmed working, I also adjusted part of the docs for `register_files` to more accurately note the behaviour of the call.


quininer:
You can also create new IoUring instances for specific test cases.

PR #346, opened 2025-07-20
Add missing ops up to Linux 6.15

silvanshade:
This PR adds support for the missing `io_uring` ops up to the current Linux 6.15 API, specifically:

- `ReadMulti`
- `RecvZc`
- `EpollWait`
- `ReadvFixed`, `WritevFixed`


silvanshade:
@quininer the workflow errors should be fixed now.

PR #349, opened 2025-07-25
Handle overflown cqes when IORING_FEAT_NODROP supported

yw-furiosa:
According to description about `IORING_FEAT_NODROP` in [man page](https://man7.org/linux/man-pages/man2/io_uring_setup.2.html) , when `IORING_FEAT_NODROP` is enabled and CQ overflows, the kernel buffers completion events internally but doesn't automatically flush them when CQ space becomes available. We must explicitly call `io_uring_enter()` to flush these buffered events, even with `SQPOLL` enabled.

I'm not sure that this code location is the best place to handle this issue. Please let me know if there is more proper location to fix this issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,4 +36,3 @@ bindgen = { version = "0.69", optional = true }
 anyhow = "1"
 socket2 = "0.5"
 slab = "0.4"
-tempfile = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,3 +36,4 @@ bindgen = { version = "0.69", optional = true }
 anyhow = "1"
 socket2 = "0.5"
 slab = "0.4"
+tempfile = "3"
--- a/io-uring-test/src/tests/sqpoll.rs
+++ b/io-uring-test/src/tests/sqpoll.rs
@@ -38,20 +38,24 @@ pub fn test_sqpoll_cq_overflow<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
 
     // Create test files for I/O operations
     let test_files = create_test_files(test_dir, num_requests);
-
-    let mut buf = vec![0u8; 1024];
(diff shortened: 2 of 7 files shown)


yw-furiosa:
I added a test case. Please review when you get a chance!

io-uring-test/src/tests/sqpoll.rs
```diff
@@ -0,0 +1,105 @@
...
+    // Create test files for I/O operations
+    let test_files = create_test_files(test_dir, num_requests);
+
+    let mut buf = vec![0u8; 1024];
+
+    let start = std::time::Instant::now();
+
+    test_files.iter().for_each(|file| {
+        let fd = file.as_raw_fd();
+        let entry = opcode::Read::new(types::Fd(fd), buf.as_mut_ptr(), buf.len() as _)
```
quininer, line 48: It's a little uncomfortable to write to the same buf at same time.

Cargo.toml
```diff
@@ -36,3 +36,4 @@ bindgen = { version = "0.69", optional = true }
 anyhow = "1"
 socket2 = "0.5"
 slab = "0.4"
+tempfile = "3"
```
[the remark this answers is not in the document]
yw-furiosa: I removed it [c594f80](https://github.com/tokio-rs/io-uring/pull/349/commits/c594f80c6bd2e8a1fa2de4083c7e5d5684c3159a)

io-uring-test/src/tests/sqpoll.rs
yw-furiosa: I changed to create buffers per Read. [019f64f](https://github.com/tokio-rs/io-uring/pull/349/commits/019f64fefa5aa6e2b40fc18cf55bb40feb23deaa)

quininer:
Thank you!

PR #355, opened 2025-08-19
Add `submitter.buf_ring_register_with_flags` and `buffer_more()` CQE flag test

### LINKED ISSUE #312: register_buf_ring missing flags
With liburing  6.12 https://github.com/axboe/liburing/wiki/What%27s-new-with-io_uring-in-6.11-and-6.12
there is the possibility to have incremental buffer consumption. It seems interesting when buffer size cannot be known ahead of time (like network). To register this feature we have to pass the flag `IOU_PBUF_RING_INC` Currently the register_buf_ring function does not accept flags. Should a new function which accept flags be created or this one updated ? 
https://docs.rs/io-uring/latest/io_uring/struct.Submitter.html#method.register_buf_ring

pinkforest:
Closes #312 

EDIT: Deprecates `submitter.buf_ring_register` in favor of `submitter.buf_ring_register_with_flags`

Also re-exports `IOU_PBUF_RING_INC` & `IOU_PBUF_RING_MMAP`

Suspect the flags were there from `IOU_PBUF_RING_MMAP` whenever that was added.

### DIFF
--- a/src/submit.rs
+++ b/src/submit.rs
@@ -584,7 +584,31 @@ impl<'a> Submitter<'a> {
     /// Developers must ensure that the `ring_addr` and its length represented by `ring_entries`
     /// are valid and will be valid until the bgid is unregistered or the ring destroyed,
     /// otherwise undefined behaviour may occur.
+    #[deprecated(note = "please use `register_buf_ring_with_flags` instead")]
     pub unsafe fn register_buf_ring(
+        &self,
+        ring_addr: u64,
+        ring_entries: u16,
+        bgid: u16,
+    ) -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


src/submit.rs
```diff
@@ -589,6 +589,7 @@ impl<'a> Submitter<'a> {
         ring_addr: u64,
         ring_entries: u16,
         bgid: u16,
+        flags: u16,
```
quininer, line 592: I don't want to break api, here we need to add a new method and mark the old one as deprecated.

quininer (changes requested):

src/submit.rs
pinkforest: Done `_with_flags` now

PR #359, opened 2025-08-25
Add description about supported bindings

mox692:
Add a description about:

* prebuilt bindings
* how to use `bindgen` feature or `io_uring_use_own_sys` cfg when you're trying to build for a target that `io-uring` does not support by default.

If there is a more appropriate place rather than `lib.rs`, let me know!

pic


quininer:
This looks good. Thank you!

PR #361, opened 2025-09-01
Update for Linux 6.16.4

### LINKED ISSUE #360: Support for linux 6.16.2
linux 6.16.2 changed the type of rw_flags from ```__kernel_rwf_t``` aka ```c_int``` aka ```i32``` to ```__u32``` aka ```u32```.

https://github.com/gregkh/linux/commit/9c0b451627cd18a6f76e9e7efadb446a3839f1ef

This breaks buliding the io-uring crate with the bindgen feature in two ways.

1. The bindings file no longer exports ```__kernel_rwf_t```
2. Rust (unlike c) does not perform implicit conversions between signed and unsigned integers.

I prepared a patch to address these issues in Debian, it is available at

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/a5092e14f5ae51e2d7b73078495439548041d33e/src/io-uring/debian/patches/linux-6.16.2.patch

silvanshade:
Fixes #360

This also adds sys files for the remaining Tier 2 targets.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -131,7 +131,7 @@ opcode! {
         offset: u64 = 0,
         /// specified for read operations, contains a bitwise OR of per-I/O flags,
         /// as described in the `preadv2(2)` man page.
-        rw_flags: types::RwFlags = 0,
+        rw_flags: i32 = 0,
         buf_group: u16 = 0
     }
 
@@ -152,7 +152,7 @@ opcode! {
         sqe.__bindgen_anon_2.addr = iovec as _;
         sqe.len = len;
         sqe.__bindgen_anon_1.off = offset;
-        sqe.__bindgen_anon_3.rw_flags = rw_flags;
+        sqe.__bindgen_anon_3.rw_flags = rw_flags as _;
         sqe.__bindgen_anon_4.buf_group = buf_group;
         Entry(sqe)
     }
@@ -170,7 +170,7 @@ opcode! {
         offset: u64 = 0,
         /// specified for write operations, contains a bitwise OR of per-I/O flags,
         /// as described in the `preadv2(2)` man page.
-        rw_flags: types::RwFlags = 0
+        rw_flags: i32 = 0
     }
 
     pub const CODE = sys::IORING_OP_WRITEV;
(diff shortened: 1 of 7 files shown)


src/sys/mod.rs
```diff
@@ -18,11 +18,15 @@ use libc::*;
 #[cfg(all(
     not(feature = "bindgen"),
     not(any(
-        target_arch = "x86_64",
         target_arch = "aarch64",
-        target_arch = "riscv64",
+        target_arch = "arm",
         target_arch = "loongarch64",
-        target_arch = "powerpc64"
+        target_arch = "powerpc",
```
quininer, line 24: I'm a bit hesitant about vendor more archs. This should be a separate PR.

silvanshade: I can move these to a separate PR.

Is there a reason to prefer not to include these while some other Tier 2 targets are already in place?

The current behavior of falling back to the `x86_64` sys module seems worse than including these given the potential for ABI issues.

src/types.rs
```diff
@@ -48,7 +48,7 @@ use std::marker::PhantomData;
 use std::num::NonZeroU32;
 use std::os::unix::io::RawFd;
 
-pub use sys::__kernel_rwf_t as RwFlags;
+pub type RwFlags = u32;
```
quininer, line 51: Since upstream already uses u32 directly, I think this can be marked as deprecated.

src/sys/mod.rs
quininer: I actually suggestion that tier 2 users specify their own sys file. that's what I do myself.

In the long term I hope to migrate to `linux-raw-sys`, but at the same time keep the `bindgen` feature.

src/opcode.rs
```diff
@@ -2305,7 +2305,7 @@ opcode! {
         ;;
         ioprio: u16 = 0,
         offset: u64 = 0,
-        rw_flags: i32 = 0,
```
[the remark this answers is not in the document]
silvanshade: Fair enough.

quininer:
Thank you!

PR #369, opened 2025-10-01
Add Entry::get_opcode().

sugarraysam:
As titled.

I found to work with the library, it's easier to accept `Entry` in functions because it can contain any of the opcode. Whereas accepting opcode directly requires generics and does not work when you need type-erasure.

Anyhow, knowing which "flavor" of entry you have is essential to do validation.


sugarraysam:
merged in #370

PR #370, opened 2025-10-14
Entry getters and setters

sugarraysam:
Adding a few convenience functions.

(1) `set_user_data` is useful for SQ push retries. Most applications working with iouring use a Slab or other indexed storage to bridge between SQ ring and CQ ring. If you retry, you have to re-index stuff, thus mutating the user_data becomes really convenient (see: https://gist.github.com/sugarraysam/01008d85a418aa71a5aac04ce2a1a623)

(2) `clear_flags` is convenience as well. API only allows "augmenting" the flags with `|=` so at least now we can reset.

(3) `get_opcode` this was hidden and is good to see if you get an opaque Entry.

### DIFF
--- a/src/squeue.rs
+++ b/src/squeue.rs
@@ -328,6 +328,13 @@ impl Entry {
         self
     }
 
+    /// Clear the submission event's [flags](Flags).
+    #[inline]
+    pub fn clear_flags(mut self) -> Entry {
+        self.0.flags = 0;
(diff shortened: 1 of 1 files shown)


sugarraysam:
But more importantly, forget the previous comment, Entry is super useful because it is type erased, so you can store any opcode seamlessly.

PR #372, opened 2025-11-01
chore: Trusted Publish

quininer:
I think this is good, it will make the version release more transparent.


quininer:
@mox692 I plan to allow only @tokio-rs/uring-core to create tags.

PR #376, opened 2025-12-15
Convert negative CQE results to io::Error

asayers:
The following text is found in many io_uring man pages:

    Note that where synchronous system calls will return -1 on failure
    and set errno to the actual error value, io_uring never uses errno.
    Instead it returns the negated errno directly in the CQE res field.

My understanding is that a negative value in the res field is _always_ a negated errno. Converting negative values to `io::Error` removes a footgun, and is often more convenient.  You can recover the original errno with `Error::raw_os_error()`.

(This is a backport of https://github.com/jordanisaacs/rustix-uring/pull/11 and https://github.com/jordanisaacs/rustix-uring/pull/15)


quininer:
I don't reject this, but it would be a break change. would you can to add another method that returns `io::Result`? like `io_result(self) -> io::Result<u32>`

asayers:
It's been a couple of weeks and I haven't thought of anything better, so I renamed it to `io_result()`.

PR #390, opened 2026-04-01
register: Add register_napi/unregister_napi methods

mtottenh:
Expose IORING_REGISTER_NAPI (opcode 27) and IORING_UNREGISTER_NAPI (opcode 28) through the Submitter API. These allow configuring per-ring NAPI busy-poll, where io_uring_enter() polls NAPI instances on the calling CPU before blocking for completions.

register_napi takes a mutable reference since the kernel writes the previous settings back to the struct. unregister_napi optionally accepts a struct to receive the current settings before disabling.

Re-export io_uring_napi and associated constants through the public types module.

Available since Linux 6.9.

### DIFF
--- a/src/submit.rs
+++ b/src/submit.rs
@@ -753,4 +753,39 @@ impl<'a> Submitter<'a> {
         )
         .map(drop)
     }
+
+    /// Register NAPI busy-poll settings on this io_uring ring.
+    ///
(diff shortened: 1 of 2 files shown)


quininer:
Please don't directly export `sys` API, as it is unstable.

rektide:
Resolved in #397!

PR #394, opened 2026-04-22
feat: add ioprio method for Send and Recv

abh1nav10:
This adds the `ioprio` setter for `Send` and `Recv` opcodes.


quininer:
Thank you!

PR #397, opened 2026-06-17
Add `register_napi`/`unregister_napi` methods

lthiery:
Expose `IORING_REGISTER_NAPI` / `IORING_UNREGISTER_NAPI` (Linux 6.9+) for per-ring NAPI busy-polling. Settings are configured through a new `Napi` wrapper and `NapiTracking` enum, keeping the unstable `sys` types and constants private rather than re-exporting them.

For the static tracking strategy, `register_napi_add_id` / `register_napi_del_id` (Linux 6.13+) manage individual NAPI ids; like the rest of the wrapper they leave precondition checks (e.g. requiring static tracking) to the kernel. Includes a test covering the register/unregister write-back and the static id ops, skipping on kernels without support.

Supercedes #390

### DIFF
--- a/io-uring-test/src/tests/register_napi.rs
+++ b/io-uring-test/src/tests/register_napi.rs
@@ -1,0 +1,74 @@
+use std::io;
+
+use io_uring::cqueue;
+use io_uring::squeue;
+use io_uring::types::{Napi, NapiTracking};
+use io_uring::IoUring;
+
+use crate::Test;
+
+pub fn test_register_napi<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> io::Result<()> {
+    require!(
+        test;
+    );
+
+    println!("test register_napi");
+
+    // Register a first configuration. The kernel writes the previous (default) settings
+    // back into `first`, so it should report a disabled busy-poll timeout.
+    let mut first = Napi::new()
+        .set_busy_poll_timeout(60)
+        .set_prefer_busy_poll(true);
+    if let Err(e) = ring.submitter().register_napi(&mut first) {
+        // IORING_REGISTER_NAPI requires Linux 6.9+.
+        if matches!(e.raw_os_error(), Some(libc::EINVAL | libc::ENOTSUP)) {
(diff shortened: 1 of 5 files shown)


io-uring-test/src/tests/register_napi.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    // Register a first configuration. The kernel writes the previous (default) settings
+    // back into `first`, so it should report a disabled busy-poll timeout.
+    let mut first = Napi::new()
+        .set_busy_poll_timeout(60)
+        .set_prefer_busy_poll(true);
+    if let Err(e) = ring.submitter().register_napi(&mut first) {
+        // IORING_REGISTER_NAPI requires Linux 6.9+.
+        if matches!(e.raw_os_error(), Some(libc::EINVAL | libc::ENOTSUP)) {
+            println!("skipping register_napi: not supported by this kernel");
```
quininer: This should be tested in `require!` https://github.com/tokio-rs/io-uring/blob/master/io-uring-test/src/utils.rs#L24

io-uring-test/src/tests/register_napi.rs
```diff
@@ -0,0 +1,74 @@
...
+        second.busy_poll_timeout(),
+        60,
+        "register_napi should return the previous busy-poll timeout"
+    );
+    assert!(
+        second.prefer_busy_poll(),
+        "register_napi should return the previous prefer_busy_poll setting"
+    );
+
+    // Read the current settings back while disabling.
```
quininer, line 57: This test is a bit simple and does not reflect the intention and effect.

but if you need it, I don’t mind merge it first.

lthiery: This test actually was useful because initially I tried to disable without passing a struct and it did not like that, so it _is_ covering some failure mode.

However, yeah it's pretty limited in scope because really doing it write needs to cross the NIC boundary and loopback doesn't do that. I'll fix the docs around the test to be more clear about the limited scope.

io-uring-test/src/tests/register_napi.rs
lthiery: Absolutely! Fixed

quininer:
Thank you!

PR #398, opened 2026-06-18
feat: add `register_ring_fd`/`unregister_ring_fd` methods

MrCroxx:
Hi. Thank you for providing this nice crate. 

I'm using `io-uring` in HPC environment, the `fget()` overhead from `io_uring_enter(2)` comsumes a lot of CPU. 

I tried to eliminate it with `IORING_REGISTER_RING_FDS` but I found this crate didn't expose the API. This PR add `register_ring_fd()` and `unregister_ring_fd()` to make `IORING_REGISTER_RING_FDS` usable. I've followed the file fd register API pattern to add the methods to `Submitter`, which is `!Send` and `!Sync` to avoid cross-thread usage.

## Benchmark

w/o ring fd register.

perf-step1b

w/ ring fd register.

perf-step1c

### DIFF
--- a/io-uring-test/src/tests/register.rs
+++ b/io-uring-test/src/tests/register.rs
@@ -68,3 +68,64 @@ pub fn test_register_files_sparse<S: squeue::EntryMarker, C: cqueue::EntryMarker
 
     Ok(())
 }
+
+pub fn test_register_ring_fd<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    _test: &Test,
+) -> anyhow::Result<()> {
+    println!("test register_ring_fd");
+
+    // register_ring_fd was introduced in kernel 5.18. Skip the test if it is not supported.
+    match ring.submitter().register_ring_fd() {
+        Ok(()) => {}
+        Err(e) if e.raw_os_error() == Some(libc::EINVAL) => {
+            println!("register_ring_fd not supported, skipping");
+            return Ok(());
+        }
+        Err(e) => return Err(anyhow::anyhow!("register_ring_fd failed: {}", e)),
+    }
+
+    // Registering twice should fail.
+    if ring.submitter().register_ring_fd().is_ok() {
+        return Err(anyhow::anyhow!(
+            "register_ring_fd should not have succeeded twice in a row"
+        ));
+    }
+
+    // Submitting now transparently goes through the registered ring index.
+    let nop = opcode::Nop::new().build().user_data(0x42).into();
+    unsafe {
+        ring.submission()
+            .push(&nop)
+            .expect("submission queue is full");
+    }
+    ring.submit_and_wait(1)?;
+    let cqes: Vec<cqueue::Entry> = ring.completion().map(Into::into).collect();
+    assert_eq!(cqes.len(), 1);
+    assert_eq!(cqes[0].user_data(), 0x42);
+    assert_eq!(cqes[0].result(), 0);
+
+    ring.submitter().unregister_ring_fd()?;
+
+    // Unregistering twice should fail.
+    if ring.submitter().unregister_ring_fd().is_ok() {
+        return Err(anyhow::anyhow!(
+            "unregister_ring_fd should not have succeeded twice in a row"
+        ));
+    }
+
+    // Submitting still works after unregistering, now via the raw file descriptor.
+    let nop = opcode::Nop::new().build().user_data(0x43).into();
+    unsafe {
+        ring.submission()
+            .push(&nop)
+            .expect("submission queue is full");
+    }
+    ring.submit_and_wait(1)?;
+    let cqes: Vec<cqueue::Entry> = ring.completion().map(Into::into).collect();
+    assert_eq!(cqes.len(), 1);
+    assert_eq!(cqes[0].user_data(), 0x43);
+    assert_eq!(cqes[0].result(), 0);
+
+    Ok(())
+}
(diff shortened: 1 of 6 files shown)


src/submit.rs
```diff
@@ -602,6 +609,63 @@ impl<'a> Submitter<'a> {
...
+    /// [`EnterFlags::REGISTERED_RING`] together with the registered index instead of the raw file
+    /// descriptor. This avoids the per-call file descriptor lookup overhead in the kernel.
+    ///
+    /// The registration is remembered by the [`IoUring`](crate::IoUring) instance and used
+    /// transparently; call [`unregister_ring_fd`](Self::unregister_ring_fd) to undo it. Calling this
+    /// while a ring fd is already registered returns an `EEXIST` error.
+    ///
+    /// Available since Linux 5.18.
+    pub fn register_ring_fd(&self) -> io::Result<()> {
+        if self.enter_ring_fd.load(atomic::Ordering::Relaxed) >= 0 {
```
quininer: I think here has condition race.

src/submit.rs
```diff
@@ -602,6 +609,63 @@ impl<'a> Submitter<'a> {
...
+    /// [`submit_and_wait`](Self::submit_and_wait)) automatically pass
+    /// [`EnterFlags::REGISTERED_RING`] together with the registered index instead of the raw file
+    /// descriptor. This avoids the per-call file descriptor lookup overhead in the kernel.
+    ///
+    /// The registration is remembered by the [`IoUring`](crate::IoUring) instance and used
+    /// transparently; call [`unregister_ring_fd`](Self::unregister_ring_fd) to undo it. Calling this
+    /// while a ring fd is already registered returns an `EEXIST` error.
+    ///
+    /// Available since Linux 5.18.
+    pub fn register_ring_fd(&self) -> io::Result<()> {
```
quininer: Is this safety?

src/lib.rs
```diff
@@ -84,6 +85,11 @@ where
     fd: OwnedFd,
     params: Parameters,
     memory: ManuallyDrop<MemoryMap>,
+
+    /// The registered ring fd index used for `io_uring_enter`, or `-1` when no ring fd is
+    /// registered. While set to a valid index (by [`Submitter::register_ring_fd`]), `enter` passes
+    /// it together with `IORING_ENTER_REGISTERED_RING` instead of the raw file descriptor.
+    enter_ring_fd: atomic::AtomicI32,
```
quininer: I don't like this a little.

quininer (changes requested):

src/lib.rs
MrCroxx: I misunderstand the behaviour of duplicated register. There is no need to store a global index.

src/submit.rs
MrCroxx: I'm not quite understand your concern. IIUC, you concern about if it is safe to use the original path after register ring fd. It is safe in that case. After ring fd is registered, you can still use the non-register call without optimization.

But I agree with your insitutaion. It's better to make it `&mut self` here, atomic here is correct but useless in most cases I suppose. Let me change it.

src/submit.rs
```diff
@@ -124,15 +126,19 @@ impl<'a> Submitter<'a> {
...
-            arg,
-            size,
-        )
-        .map(|res| res as _)
+
+        // If a ring fd has been registered with [`register_ring_fd`](Self::register_ring_fd),
+        // `enter_ring_fd` holds its index (otherwise `-1`); pass it together with
+        // `IORING_ENTER_REGISTERED_RING` instead of the raw file descriptor to avoid the per-call
+        // fd lookup in the kernel.
+        let enter_ring_fd = self.enter_ring_fd;
```
quininer, line 154: Do we need to do this for `register::execute`?

io-uring-test/src/tests/register.rs
```diff
@@ -68,3 +68,118 @@ pub fn test_register_files_sparse<S: squeue::EntryMarker, C: cqueue::EntryMarker
...
+    _test: &Test,
+) -> anyhow::Result<()> {
+    println!("test register_ring_fd");
+
+    // register_ring_fd was introduced in kernel 5.18. Skip the test if it is not supported.
+    let mut submitter = ring.submitter();
+    match submitter.register_ring_fd() {
+        Ok(()) => {}
+        Err(e) if e.raw_os_error() == Some(libc::EINVAL) => {
+            println!("register_ring_fd not supported, skipping");
```
quininer: Use `require!` see https://github.com/tokio-rs/io-uring/blob/master/io-uring-test/src/utils.rs#L4

MrCroxx: Got it. However, I've checked the methods. I'm afraid there isn't a OPCODE that indicates if `register_ring_fd` is supported. The best way I come up with is to call `register_ring_fd` on a not-related ring and check the return value. Then, run the following tests. I've already tried it with the new commit. Please take a look if it satifies your needs.

src/submit.rs
```diff
@@ -662,7 +703,7 @@ impl<'a> Submitter<'a> {
     /// Available since Linux 6.9.
     pub fn register_napi(&self, napi: &mut Napi) -> io::Result<()> {
         execute(
-            self.fd.as_raw_fd(),
+            RegisterRing::RawFd(self.fd.as_raw_fd()),
```
quininer: doesn't `register_napi` need to check `enter_ring_fd`?

MrCroxx: Checked man pages again. You are right! Let me fix it.

quininer:
Thank you!

MrCroxx:
> Thank you!

Also thank you for your guide. 🥰

PR #399, opened 2026-06-22
Add `Parameters::is_feature_no_iowait` accessor

hubcio:
The IORING_FEAT_NO_IOWAIT bit is already in the generated bindings but had
no accessor, so callers had to probe IORING_ENTER_NO_IOWAIT support via an
EINVAL round-trip. Expose it alongside the other is_feature_* methods.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -654,6 +654,18 @@ impl Parameters {
         self.0.features & sys::IORING_FEAT_MIN_TIMEOUT != 0
     }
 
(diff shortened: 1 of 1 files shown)


Berrysoft:
Hello, @quininer ! Could you publish a new release containing this PR? We need it to optimize our runtime's performance.

quininer:
@Berrysoft  I will release a version on weekend. By the way, if you want request release a patch version, you can open a PR for this. :)

PR #400, opened 2026-06-22
io-uring-test: fix flaky epoll_wait_race short-read assertion

lthiery:
The reader buffer in `prune_read` was `[0u8; 16]`, which is not a multiple of `DATA.len()` (3). When the writer thread gets ahead and a pipe holds more than 16 bytes, a single read stops mid-`DATA`, leaving a partial chunk in the pipe. Level-triggered epoll then re-reports the pipe readable and the following read returns that 2-byte remainder, so `assert!(len >= DATA.len())` fails.

It is timing dependent (the reader must fall ~6 writes behind on a pipe), which is why it only surfaced intermittently on [the latest-kernel CI job](https://github.com/tokio-rs/io-uring/actions/runs/27969949046/job/82773345120). Sizing the buffer to a multiple of `DATA.len()` keeps every read aligned to whole `DATA` chunks, so reads never split a write.

### DIFF
--- a/io-uring-test/src/tests/epoll.rs
+++ b/io-uring-test/src/tests/epoll.rs
@@ -289,7 +289,12 @@ pub fn test_race<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
         nr: usize,
     ) -> anyhow::Result<()> {
         for rx_idx in events.iter().take(nr) {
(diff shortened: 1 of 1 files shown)


quininer:
Thank you!

PR #402, opened 2026-07-10
feat: add range support (offset/len) for IORING_OP_FSYNC

### LINKED ISSUE #389: Missing range support for IORING_OP_FSYNC
The [man page for `IORING_OP_FSYNC`](https://man7.org/linux/man-pages/man2/io_uring_enter.2.html) currently states:

> Optionally off and len can be used to specify a range within the file to be synced rather than syncing the entire file, which is the default behavior.

Would be great if this was also supported here.  
Judging from the code, this might be all that's needed:
```patch
Subject: [PATCH] IORING_OP_FSYNC: add file range support (off, len)
---
Index: src/opcode.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/opcode.rs b/src/opcode.rs
--- a/src/opcode.rs	(revision 1c0c4116ad194592cd4d0ed1596a0000371519b1)
+++ b/src/opcode.rs	(date 1774897550886)
@@ -195,18 +195,24 @@
 }
 
 opcode! {
-    /// File sync, equivalent to `fsync(2)`.
+    /// File sync, See also `fsync(2)`.
     ///
+    /// Optionally `off` and `len` can be used to specify a range within the file to be synced rather
+    /// than syncing the entire file, which is the default behavior.
     /// Note that, while I/O is initiated in the order in which it appears in the submission queue,
     /// completions are unordered. For example, an application which places a write I/O followed by
     /// an fsync in the submission queue cannot expect the fsync to apply to the write. The two
     /// operations execute in parallel, so the fsync may complete before the write is issued to the
     /// storage. The same is also true for previously issued writes that have not completed prior to
-    /// the fsync.
+    /// the fsync. To enforce ordering one may utilize linked SQEs, `IOSQE_IO_DRAIN` or wait for the
+    /// arrival of CQEs of requests which have to be ordered before a given request before
+    /// submitting its SQE.
     #[derive(Debug)]
     pub struct Fsync {
         fd: { impl sealed::UseFixed },
         ;;
+        offset: u64 = 0,
+        len: u32 = 0,
         /// The `flags` bit mask may contain either 0, for a normal file integrity sync,
         /// or [types::FsyncFlags::DATASYNC] to provide data sync only semantics.
         /// See the descriptions of `O_SYNC` and `O_DSYNC` in the `open(2)` manual page for more information.
@@ -216,11 +222,13 @@
     pub const CODE = sys::IORING_OP_FSYNC;
 
     pub fn build(self) -> Entry {
-        let Fsync { fd, flags } = self;
+        let Fsync { fd, offset, len, flags } = self;
 
         let mut sqe = sqe_zeroed();
         sqe.opcode = Self::CODE;
         assign_fd!(sqe.fd = fd);
+        sqe.len = len;
+        sqe.__bindgen_anon_1.off = offset;
         sqe.__bindgen_anon_3.fsync_flags = flags.bits();
         Entry(sqe)
     }
```

But I didn't test (for lack of knowledge on how) whether this has any effect. At least it didn't explode for my tests: the return code from the CQE always returns 0 for me, including for `.offset(u64::MAX).len(u32::MAX)`, `.offset(0).len(u32::MAX)`, `.offset(u64::MAX).len(0)` and `.offset(0).len(0)`.

See also https://github.com/axboe/liburing/issues/990.

RajaBabu15:
## Summary

Adds optional `offset` and `len` fields to `opcode::Fsync`, matching the kernel `IORING_OP_FSYNC` interface described in the [io_uring_enter(2)](https://man7.org/linux/man-pages/man2/io_uring_enter.2.html) man page:

> Optionally off and len can be used to specify a range within the file to be synced rather than syncing the entire file, which is the default behavior.

- Defaults remain `offset = 0`, `len = 0` (whole-file sync), so existing callers are unchanged.
- SQE fields are wired the same way as other range-aware opcodes (`Write`, `SyncFileRange`, etc.).
- Docs note ordering caveats and how to enforce ordering (linked SQEs / `IOSQE_IO_DRAIN` / waiting for CQEs).

Fixes #389

## Test plan

- [x] Unit tests: `opcode::tests::fsync_default_range_is_zero`, `fsync_range_and_flags_are_wired` (SQE field wiring)
- [x] Integration: `test_file_fsync_range` in `io-uring-test` (runs under Linux; verified in Docker LinuxKit 6.12 with Rust 1.95)
- [x] `cargo check` / `cargo test -p io-uring --lib -- fsync` on Linux
- [ ] CI on this PR

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -195,18 +195,30 @@ opcode! {
 }
 
 opcode! {
-    /// File sync, equivalent to `fsync(2)`.
+    /// File sync. See also `fsync(2)`.
+    ///
+    /// Optionally [`offset`](Self::offset) and [`len`](Self::len) can be used to specify a range
+    /// within the file to be synced rather than syncing the entire file, which is the default
+    /// behavior.
     ///
     /// Note that, while I/O is initiated in the order in which it appears in the submission queue,
     /// completions are unordered. For example, an application which places a write I/O followed by
     /// an fsync in the submission queue cannot expect the fsync to apply to the write. The two
     /// operations execute in parallel, so the fsync may complete before the write is issued to the
     /// storage. The same is also true for previously issued writes that have not completed prior to
-    /// the fsync.
+    /// the fsync. To enforce ordering one may utilize linked SQEs, `IOSQE_IO_DRAIN` or wait for the
+    /// arrival of CQEs of requests which have to be ordered before a given request before
+    /// submitting its SQE.
     #[derive(Debug)]
     pub struct Fsync {
         fd: { impl sealed::UseFixed },
         ;;
+        /// Offset within the file to start the sync range. Together with [`len`](Self::len), this
+        /// selects a byte range; when both are zero (the default), the entire file is synced.
+        offset: u64 = 0,
+        /// Length of the range to sync, in bytes. When zero (the default) with
+        /// [`offset`](Self::offset) also zero, the entire file is synced.
+        len: u32 = 0,
         /// The `flags` bit mask may contain either 0, for a normal file integrity sync,
         /// or [types::FsyncFlags::DATASYNC] to provide data sync only semantics.
         /// See the descriptions of `O_SYNC` and `O_DSYNC` in the `open(2)` manual page for more information.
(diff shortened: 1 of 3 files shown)


src/opcode.rs
```diff
@@ -2406,3 +2425,38 @@ opcode! {
         Entry(sqe)
     }
 }
+
+#[cfg(test)]
+mod tests {
```
quininer: We don’t write tests here.

quininer:
Thank you!

PR #403, opened 2026-07-29
Expose user data through queue entry traits

Orkking2:
Runtimes (e.g., [my nosv package](https://github.com/Orkking2/nosv)) wrapping io-uring may replace an SQE's application-supplied `user_data` with an internal token used to associate completions with runtime state. To preserve the public API's behavior, the runtime must be able to save the original value and restore it on the CQE before returning the completion to the application.

This change:

- exposes user-data access through `squeue::EntryMarker`
- adds the missing `Entry128::get_user_data`
- exposes user-data access and replacement through `cqueue::EntryMarker`
- adds mutable CQE setters for `Entry` and `Entry32`
- tests the save/replace/restore behavior for both entry widths

Validation:

- `cargo test --workspace`
- `cargo clippy -p io-uring --lib`
- formatting check


quininer:
Thank you!

PR #404, opened 2026-08-02
Expose whether the completion queue has entries pending

dahankzter:
## Motivation

A thread-per-core runtime needs to know whether the kernel has posted a completion **on its hot path**, without entering the kernel, taking a lock, or borrowing the queue.

Concretely, in [glommio](https://github.com/glommio/glommio) this check runs on every scheduler iteration and every cooperative yield point, from arbitrary depth inside task execution — including from user code. The ring itself is behind a `RefCell` and may already be borrowed at that moment, so **reaching the queue costs far more than the comparison does**.

Nothing on the current public surface allows it:

- `CompletionQueue` keeps `head` and `tail` private, exposing only `sync`, `overflow`, `eventfd_disabled`, `capacity`, `is_empty`, `is_full`, `fill` and iteration — all of which require the borrow.
- They can't be reconstructed from outside either: `IoUring::params()` returns a `Parameters` whose public API is six `is_*` predicates, so `cq_off` isn't reachable.

glommio maintains a vendored copy of the abandoned `iou`/`uring-sys` crates partly because of this. This is what lets that copy — ~3,000 lines and ~100 `unsafe` occurrences — be deleted in favour of this crate.

## What this adds

```rust
/// Obtained from `CompletionQueue::status`. Borrows nothing, so it can be
/// captured once at startup and consulted from anywhere afterwards.
pub struct CompletionStatus { /* head, tail */ }

impl CompletionStatus {
    pub fn is_empty(&self) -> bool;
}

impl<E: EntryMarker> CompletionQueue<'_, E> {
    /// # Safety
    /// The returned value must not outlive the ring.
    pub unsafe fn status(&self) -> CompletionStatus;
}
```

The `unsafe` sits at a single construction site at startup instead of on a path executed millions of times a second. On the consumer side the check becomes `!self.status.is_empty()`, with no `unsafe` at all.

**Staleness is documented on the type.** The answer is stale as soon as it is returned — the kernel may post a completion immediately after the load. That is inherent to any check which does not enter the kernel, and it is worth stating plainly because a safe-looking `is_empty()` hides it more effectively than raw pointers would. It is right for hints such as "should I stop and poll?", where being one iteration late costs nothing; it is not a synchronization primitive. The existing `CompletionQueue::is_empty` carries the same caveat.

## Scope

One method, one small type, one test. No behaviour change, no new state, nothing else touched. `io-uring-test` passes across all four ring configurations, and the new test covers empty, after a completion is posted, and after it is consumed.

Naming is yours to pick — `status()`/`is_empty()` mirrors the existing `CompletionQueue::is_empty`, but `need_preempt`, `has_completions` or `is_ready` are all fine by me, as is hanging it off `IoUring` directly if reaching through `completion()` is awkward.

*(Earlier revisions of this PR exposed the raw head/tail pointers instead; reworked into this shape at @quininer's suggestion.)*

### DIFF
--- a/io-uring-test/src/main.rs
+++ b/io-uring-test/src/main.rs
@@ -73,6 +73,7 @@ fn test<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
     tests::queue::test_nop(&mut ring, &test)?;
     tests::queue::test_setup_no_sqarray(&mut ring, &test)?;
     tests::queue::test_queue_split(&mut ring, &test)?;
+    tests::queue::test_completion_status(&mut ring, &test)?;
     tests::queue::test_debug_print(&mut ring, &test)?;
     tests::queue::test_msg_ring_data(&mut ring, &test)?;
     tests::queue::test_msg_ring_send_fd(&mut ring, &test)?;
--- a/io-uring-test/src/tests/queue.rs
+++ b/io-uring-test/src/tests/queue.rs
@@ -341,3 +341,38 @@ pub fn test_msg_ring_send_fd<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
 
     Ok(())
 }
+
+pub fn test_completion_status<S: squeue::EntryMarker, C: cqueue::EntryMarker>(
+    ring: &mut IoUring<S, C>,
+    test: &Test,
+) -> anyhow::Result<()> {
+    require! {
+        test;
+    }
+
+    println!("test completion_status");
+
+    // SAFETY: the status is dropped well before `ring` is.
+    let status = unsafe { ring.completion().status() };
+
+    assert!(status.is_empty());
+
+    let nop_e = opcode::Nop::new().build().user_data(0x43).into();
+    unsafe {
+        ring.submission().push(&nop_e).expect("queue is full");
+    }
+    ring.submit_and_wait(1)?;
+
+    // The status observes the kernel's tail directly, so it reports the
+    // completion without the queue having been synchronized.
+    assert!(!status.is_empty());
+
+    let cqes: Vec<cqueue::Entry> = ring.completion().map(Into::into).collect();
+    assert_eq!(cqes.len(), 1);
+    assert_eq!(cqes[0].user_data(), 0x43);
+
+    // ... and consuming it moves the head back level with the tail.
+    assert!(status.is_empty());
+
+    Ok(())
+}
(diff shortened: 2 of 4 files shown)


quininer:
Can we expose a `need_preempt` method?

dahankzter:
Would prefer a method too. One constraint that shapes it: the check can't borrow the queue. In glommio the ring sits behind a `RefCell` and `need_preempt()` is called from arbitrary depth inside task execution, sometimes while the reactor is already borrowed — so reaching the queue is the expensive part, not the comparison.

That means a handle grabbed once, rather than a method on `CompletionQueue`:

```rust
pub struct CompletionStatus {
    head: *const AtomicU32,
    tail: *const AtomicU32,
}

impl CompletionStatus {
    pub fn is_empty(&self) -> bool { ... }
}

impl CompletionQueue<'_, E> {
    /// # Safety
    /// The returned value must not outlive the ring.
    pub unsafe fn status(&self) -> CompletionStatus { ... }
}
```

I'd rather land that than the pointers — the `unsafe` ends up at one construction site instead of on a path that runs millions of times a second.

One thing worth saying plainly: the answer is stale the moment you have it, since the kernel can post a completion immediately after the load. That's equally true of my current patch, but a safe-looking `is_empty()` probably hides it better than two raw pointers do, so it deserves a line in the docs either way. Fine for preemption — a check that's late by one iteration costs nothing — but it's not a synchronization primitive.

On the name: `need_preempt` is glommio's word for what it does with the fact; the crate only knows the CQ is non-empty. Happy either way, or `has_completions` / `is_ready`.

Say which you prefer and I'll push it.

dahankzter:
Any thoughts on this @quininer ? It's a small but pretty valuable little addition that should disturb anything else.

quininer:
That looks good. I’ll deal with it on weekend.

dahankzter:
Pushed, so it's ready to review rather than write. The PR is now a single commit implementing the handle as sketched:

- `CompletionStatus`, owned, holding the two indices — borrows nothing, so it can be captured at startup and consulted from anywhere.
- `unsafe CompletionQueue::status()` to construct it; the only invariant is that it must not outlive the ring.
- Safe `is_empty()`, plus a `Debug` impl.
- Staleness documented on the type, as promised, noting `is_empty()` on the queue itself carries the same caveat.
- A test in `io-uring-test` covering empty, after a completion is posted, and after it is consumed.

I went with `status()`/`is_empty()` to match the existing `CompletionQueue::is_empty`, but the naming is yours to pick — say the word and I'll rename.

No rush on the weekend; force-pushed over the pointer version, so there's nothing stale to read.

quininer:
Thank you!

